PR #20, opened 2018-08-01
Add Spain provinces

nickpeihl:
Adding Spain provinces and communities vector layer.

CI builds will fail until #17 is merged.


nickpeihl:
Superceded by #34.

PR #23, opened 2018-08-15
Headers must be defined before cp for gsutil.

nickpeihl:


build.sh
```diff
@@ -108,6 +108,6 @@ else
 
     # Copy files
     echo "Copying $PWD/dist/* to gs://$TARGET_BUCKET"
-    gsutil -m cp -r -a public-read -Z -h "Content-Type:application/json" $PWD/dist/* "gs://$TARGET_BUCKET"
+    gsutil -m -h "Content-Type:application/json" cp -r -a public-read -Z $PWD/dist/* "gs://$TARGET_BUCKET"
```
nyurik: i would recommend `-h "Cache-Control:public, max-age=3600"` (probably with a different time?), plus remove the `-a public-read` -- shouldn't be needed any more

Issue #27, opened 2018-08-21
Attribution should be presented as array of values, not | delimited


thomasneirynck:
Currently, attribution is pipe delimited.

Future versions should just model this as an array.

```
{
"layers": [
{
"attribution": "[Made with NaturalEarth](http://www.naturalearthdata.com/about/terms-of-use) | [Elastic Maps Service](https://www.elastic.co/elastic-maps-service)",
"name": "World Countries",
```

Will require corresponding change in Kibana

nickpeihl:
Fixed in #60

PR #28, opened 2018-08-21
[chore] Fix OpenStreetMap attribution link

nickpeihl:
The OpenStreetMap attribution link should match how it's done in the tiles manifest to avoid duplicate attributions in Kibana region maps. The duplicate EMS attributions should be fixed by resolving the issue in https://github.com/elastic/kibana/pull/22003#issuecomment-414770867. 
screen shot 2018-08-21 at 11 22 41 am

### DIFF
--- a/sources/au/states.hjson
+++ b/sources/au/states.hjson
@@ -4,7 +4,7 @@
     type: http
     note:  States of Australia
     data: https://sophox.org/regions/geojson.json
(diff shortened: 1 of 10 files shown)


thomasneirynck:
cf. https://github.com/elastic/ems-file-service/issues/27 (semi related).

PR #32, opened 2018-08-24
United Kingdom subdivisions

nickpeihl:
This layer includes all United Kingdom subdivisions listed in the [ISO 3166-2 standard](https://www.iso.org/obp/ui/#iso:code:3166:GB) with the following exceptions:
* GB-GBN - Great Britain
* GB-NIR - Northern Ireland
* GB-EAW - England and Wales
* GB-UKM - United Kingdom

The exceptions are separate nations so I don't think it makes sense to add them to this list.

Since boundary data is not complete in OpenStreetMap, we must manually process data from Ordnance Survey. I've included instructions on how that is done.


sources/uk/subdivisions.hjson
```diff
@@ -0,0 +1,37 @@
...
+            desc: Subdivision name (English)
+        }
+    ]
+    name: uk_subdivisions
+    humanReadableName: United Kingdom Subdivisions
+    conform: {
+        type: geojson
+    }
+    createdAt: 2018-08-24T17:49:30.145142000
+    filename: uk_subdivisions_v1.json
```
nickpeihl: Needs an `id` field generated by the current production EMS architecture.

PR #34, opened 2018-08-28
Spain subdivisions

nickpeihl:
This adds all Spain subdivisions. Please use this PR instead of #20 which was broken amid repo name changes, etc.


sources/es/provinces.hjson
```diff
@@ -0,0 +1,71 @@
+{
+    versions: '>=1'
+    type: http
+    production: true
+    note: Spain communities and provinces (some provinces missing, need to be added manually)
```
thomasneirynck: this note is a little ambiguous (?). does this mean the data `has already` been added manually in the geojson, or that they are missing from the geojson and `still` need to be added manually.

PR #39, opened 2018-09-10
Remove "Canton" qualifier from Switzerland labels

nickpeihl:
This removes all variations of "Canton of" from the label fields in the Switzerland cantons dataset. 

As noted by @dschneiter, including the word Canton in the label for Switzerland cantons is roughly the equivalent of saying "State of California" or "State of New York" in the US. 

Currently, label fields can be used in region map visualizations, although we [try to discourage that](https://www.elastic.co/blog/new-country-subdivision-vectors-in-the-elastic-maps-service).  This PR has the potential to break existing region map visualizations where the label is used as a join field. However, I expect any fallout from this change to be almost nil for a few reasons.

1) The Switzerland layer is very new and probably has not been included in many visualizations yet.
2) The labels currently in production for Switzerland are likely difficult to join to data because of the verbosity (ex. `Canton of Valais`).


nickpeihl:
Diffs on this JSON is impossible to read in GitHub.

My suggestion is to check out my branch locally and run 
`git diff --color-words=. master`

PR #48, opened 2018-09-13
Add field types to sources

### LINKED ISSUE #47: Seperate `id` and `property` fields in vector manifest
We want to discourage users from joining on label and name fields. Instead we plan to assign each field as type `id` or `property`. Property fields should not show up as Join Fields in Kibana by default. Enabling the use of property fields as a join field could be configured as an option in `Kibana.yml`.

nickpeihl:
This PR adds field types to the data sources. A field can have either type `id` or `property`. In the future we want to discourage the use of `property` fields as Join Fields in Kibana region maps and the GIS app. `id` fields are meant to be used for Join Fields instead as they are usually distinct defined by a standard such as the ISO 3166 country codes.  

This has no effect on `v1` and `v2` manifests.

Fixes #47.

### DIFF
--- a/schema/source_schema.json
+++ b/schema/source_schema.json
@@ -41,7 +41,8 @@
     "fieldMapping": {
       "description":
         "Specifies map source field names or properties to elastic map service properties",
-      "type": "array"
+      "type": "array",
+      "items": { "$ref": "#/definitions/fieldMap" }
     },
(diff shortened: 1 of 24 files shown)


nickpeihl:
This PR has been superseded by #60.

PR #55, opened 2018-10-22
Make GeoJSON geometry compatible with Elasticsearch

nickpeihl:
The [`geo-shape`](https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-shape.html) datatype in Elasticsearch requires geometries to conform to the [OGC Simple Feature Access specification](http://www.opengeospatial.org/standards/sfa).  But geospatial data transformation libraries such as [mapshaper](https://mapshaper.org/) and [GDAL](https://www.gdal.org/) do not ensure that geometries are valid or simple.

This PR adds a test to check that all geometries are simple and valid. In most cases, we can fix invalid and non-simple geometries with a zero-buffer (`geometry.buffer(0)`). So this PR also adds a utility script (`scripts/clean-geom.js`) that checks for and fixes invalid and non-simple geometries.


test/source.js
```diff
@@ -29,10 +30,27 @@ function testAllSources(validate) {
...
+        validateGeometry(geojson, t);
+      });
+    }
   });
 }
+
+function validateGeometry(geojson, t) {
+  const reader = new jsts.io.GeoJSONReader();
+  const fc = reader.read(geojson);
+  t.ok(fc.features.every(feat => {
```
nyurik: minor simplification here and the same below.
```suggestion
  t.ok(fc.features.every(feat => feat.geometry.isSimple()
```

PR #58, opened 2018-11-06
Add localized layer names to sources

nickpeihl:
This PR adds localized layer names from Wikidata/Wikipedia to layer sources. Translations have been manually spot-checked for consistency. Some Wikidata translations lack specificity to a specific country such as `Provins` rather than `Kinas provinser`. The specific name is then manually copied from the [associated Wikipedia page](https://sv.wikipedia.org/wiki/Kinas_provinser) (usually the title) whenever possible. 

Localization efforts in Kibana v6.6 only target English (en), German (de), and Chinese (zh) so the schema validation only [checks for those translations](https://github.com/nickpeihl/ems-file-service/blob/596e872bd5cc25a7e5e818bdf43601a8c8a18184/schema/source_schema.json#L102-L106). This requirement is likely to change in future releases.

This PR does not change the catalogue or vector manifests currently in production. This only lays the groundwork for localizing the manifests in future releases.


nyurik:
@nickpeihl after looking through translations, I think we may have to adjust it a little bit. Having some names as plural and other as singular makes it look very unclean in other languages.  So instead, lets try to come up with a "wrapper phrase", e.g.  `region type: $1` (e.g. `region type: Australian state`), where the `$1` would be the localized string as it comes from WD.  This way our translators can translate the wrapper phrase, and place the $1 in the proper position according to the rules of the language, and we can continue generating all these names from WD.  What do you think?

nickpeihl:
> @nickpeihl after looking through translations, I think we may have to adjust it a little bit. Having some names as plural and other as singular makes it look very unclean in other languages. So instead, lets try to come up with a "wrapper phrase", e.g. `region type: $1` (e.g. `region type: Australian state`), where the `$1` would be the localized string as it comes from WD. This way our translators can translate the wrapper phrase, and place the $1 in the proper position according to the rules of the language, and we can continue generating all these names from WD. What do you think?

@nyurik Our current precedent for the drop-down in Kibana is showing plural labels in English for the regions, (e.g. Australia States, World Countries, etc).

screen shot 2018-11-06 at 10 43 35 am

The translations from Wikidata (in most cases) are singular (e.g. State of Australia, Country, etc). If I understand correctly, you suggest we always store the singular translation for all languages (including English) in the internal source metadata. Then do we go against our current precedent and always show singular labels in the drop down box in Kibana? Or do we have some way of creating a plural label for the manifest and Kibana?

nyurik:
As mentioned in the zoom syncup -- I think the `Vector map` label for the dropdown should be changed to something relevant for the singular items, e.g. `Map unit`, `Region type`, `Region unit`, ...   Naming TBD :)

nickpeihl:
This PR has been superseded by #60.

Issue #56, opened 2018-10-24
[DISCUSS] EMS Localization


nickpeihl:
Related to https://github.com/elastic/kibana/issues/23407

Localization in EMS will require changes to the EMS manifests and Kibana.

Region maps in Kibana will need to access the `locale` setting in `kibana.yml` and use that to determine which vector manifest to choose from the catalogue manifest ([modifying this function](https://github.com/elastic/kibana/blob/master/src/ui/public/vis/map/service_settings.js#L89-L104)).

## Catalogue manifest schema

Here's a first draft of how I think the catalogue manifest would appear for the vector layers. We would generate different vector manifests for each locale in the `i18n` dictionary and fall back to the top level `manifest` if necessary.. The list of locales should match the locales in the [kbn-i18n package ](https://github.com/elastic/kibana/blob/f522b31ac986d10e38f487055f70c1e8350ee741/packages/kbn-i18n/src/core/locales.js). Note, I haven't changed anything in the `tiles_v2` element, but that could change, too, depending on how we handle basemap internationalization.

```json
{
  "services": [
    {
      "id": "tiles_v2",
      "name": "Elastic Maps Tile Service",
      "manifest": "https://tiles.maps.elastic.co/v2/manifest",
      "type": "tms"
    },
    {
      "id": "geo_layers",
      "name": "Elastic Maps Vector Service",
      "manifest": "https://vector.maps.elastic.co/v6.6/en-US/manifest",
      "type": "file",
      "i18n": { 
        "af": {
          "manifest": "https://vector.maps.elastic.co/v6.6/af/manifest"
        },
        ...,
        "es": {
          "manifest": "https://vector.maps.elastic.co/v6.6/es/manifest"
        },
        ...,
      }
    }
  ]
}
```

## Vector manifests schema 

Here is an example of how I assume the Spanish locale vector manifest would appear (https://vector.maps.elastic.co/v6.6/es/manifest). The top level `name` property in each layer must remain the same to avoid breaking visualizations but a `description` property is added which could be used as the `label` in the select box instead of the `name` property. This would be similar to how the [`description` property in the `fields` is used](https://github.com/elastic/kibana/blob/master/src/core_plugins/region_map/public/region_map_vis_params.html#L36-L48). 

```json
{
  "layers": [
    {
      "attribution": "[Made with NaturalEarth](http://www.naturalearthdata.com/about/terms-of-use)|[Elastic Maps Service](https://www.elastic.co/elastic-maps-service)",
[...]

PR #60, opened 2018-11-15
EMS Version 6.6

nickpeihl:
WIP - Please do not merge.

EMS v6.6 adds localized layer names and field labels based on translations available in Wikidata. The vector manifest is changed based on discussions in #56.  

TODO
* [x] Update documentation
* [x] Review changes needed to Kibana

Fixes #58 #48


schema/source_schema.json
```diff
@@ -96,6 +98,16 @@
...
   "definitions": {
+    "i18n-attribution": {
+      "type": "array",
+      "items": [
+        { "type": "string" }
+      ]
+    },
+    "i18n": {
+      "type": "object",
+      "required": ["en", "de", "zh"]
```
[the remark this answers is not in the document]
nickpeihl: I think it makes sense we only require `en` here instead since we might not always have translations available for each version.

scripts/generate-manifest.js
```diff
@@ -53,38 +53,49 @@ function generateCatalogueManifest(opts) {
...
  * @param {string} [opts.hostname=`${constants.VECTOR_STAGING_HOST}`] - Hostname for files in manifest
+ * @param {Object} [opts.fieldInfo=null] - Field metadata
  */
 function generateVectorManifest(sources, opts) {
   opts = {
     version: 'v0',
     production: false,
-    hostname: constants.VECTOR_STAGING_HOST, ...opts,
+    hostname: constants.VECTOR_STAGING_HOST,
+    fieldInfo: null, ...opts,
```
nyurik: opts shouldn't be on the same line - confusing
```suggestion
    fieldInfo: null,
    ...opts,
```

scripts/generate-manifest.js
```diff
@@ -128,3 +139,49 @@ function manifestLayerV2(data, hostname) {
...
+}
+
+function getDefaultFormat(emsFormats) {
+  return emsFormats.filter(format => format.default);
+}
+
+function getFieldLabels(fieldName, fieldInfo) {
+  if (fieldName.startsWith('label_') && _.get(fieldInfo, 'name.i18n')) {
+    const lang = fieldName.replace('label_', '');
+    const labels = _.mapValues(fieldInfo.name.i18n, (label) => {
```
nyurik:
```suggestion
    const labels = _.mapValues(fieldInfo.name.i18n, label => `${label} (${lang})`);
```

Note that `${label} (${lang})` is not a good translation string -- because in some languages it would look horrible, especially in RTL.  Usually, the formatting string itself is also stored in some translation system, so that users can provide locale-specific formatting, e.g. `label [lang]` or something funkier. This probably should be a separate patch though, once the translation method is established.

scripts/generate-manifest.js
```diff
@@ -128,3 +139,49 @@ function manifestLayerV2(data, hostname) {
...
+      return `${label} (${lang})`;
+    });
+    return labels;
+  } else if (_.get(fieldInfo, `${fieldName}.i18n`)) {
+    return fieldInfo[fieldName].i18n;
+  } else return;
```
nyurik: Easy to miss that return
```suggestion
  } else {
    return;
  }
```

scripts/generate-manifest.js
```diff
@@ -128,3 +139,49 @@ function manifestLayerV2(data, hostname) {
...
+
+function getDefaultFormat(emsFormats) {
+  return emsFormats.filter(format => format.default);
+}
+
+function getFieldLabels(fieldName, fieldInfo) {
+  if (fieldName.startsWith('label_') && _.get(fieldInfo, 'name.i18n')) {
+    const lang = fieldName.replace('label_', '');
+    const labels = _.mapValues(fieldInfo.name.i18n, (label) => {
+      return `${label} (${lang})`;
```
nyurik:
```suggestion
```

scripts/generate-manifest.js
```diff
@@ -128,3 +139,49 @@ function manifestLayerV2(data, hostname) {
...
+    });
```
nyurik:
```suggestion
```

scripts/generate-manifest.js
```diff
@@ -128,3 +139,49 @@ function manifestLayerV2(data, hostname) {
...
+  } else if (_.get(fieldInfo, `${fieldName}.i18n`)) {
+    return fieldInfo[fieldName].i18n;
+  } else return;
+}
+
+function getAttributionString(attr) {
+  if (!attr.label) {
+    throw new Error(`Attribution sources must have a 'label' property`);
+  }
+  return attr.url ? `[${attr.label.en}](${attr.url.en})` : `${attr.label.en}`;
```
nyurik: I thought we were switching to the individual fields. Is this for the older versions?

nickpeihl: Yes, that's for the v1 and v2 manifests. I can add documentation here.

scripts/generate-manifest.js
thomasneirynck: Agreed we could patch this separately. It wouldn't impact 6.6 users as this is just a placeholder for now. fwiw, I do agree this will be a longer term issue that we need to resolve (and I'm the wrong person to give meaningful feedback on this topic).

nickpeihl: As we start to support additional languages we should see how this is done in Kibana and patch the manifest as necessary. 👍

PR #69, opened 2018-12-07
Field Tests and Extend Geometry Tests to TopoJSON

nickpeihl:
Several things are happening in this PR. 

1. A new test verifies the fields defined in the manifests actually exist in the vector dataset.
1. Expands the simple and valid geometry tests to TopoJSON files. This ensures that TopoJSON files converted to GeoJSON can be ingested into Elasticsearch.
1. As a result of the new geometry tests, several null or invisible geometries in the [production USA Zip Codes dataset](https://maps.elastic.co/v2/index.html#file/USA%20zip%20codes) have been detected and subsequently deleted. Since these geometries can not be seen in production there should be no adverse affect for users. The deleted zip codes are:

- 03853
- 49831
- 49894
- 50465
- 51355
- 55107
- 96006

4. Additional geometry errors in the USA zip codes dataset have been repaired by `geometry.buffer(0)`. A very small (~10) set of features had to be repaired manually.
4. The new USA Counties dataset in #65 is included in this PR. So this PR should be merged after #65.
4. The USA zip codes dataset is sorted and prettified in the repo for better diffing in the future. As of #64 vector datasets will be automatically minified for production usage.


nickpeihl:
This should really be split into two PRs. 

1) Geometry tests for topojson
2) Vector field tests for all formats

PR #70, opened 2018-12-28
Add Italy provinces

nickpeihl:
This PR adds 103 Italian provinces (and province-equivalents). This should reflect the most current provincial boundaries. 

Italy's provinces have changed several times over the past few years. Some have been abolished and boundaries are no longer available. The [ISO Country Codes](https://www.iso.org/obp/ui/#iso:code:3166:IT) standard has not been updated to reflect all changes. This dataset includes the ISTAT ID as another Join field. 

Please review the notes in the included [README file](https://github.com/elastic/ems-file-service/compare/master...nickpeihl:italy?expand=1#diff-add7d23de4a0fd3f3dd2c555ff838624).


templates/source_template.hjson
```diff
@@ -61,7 +61,8 @@
                     OPTIONAL { ?id rdfs:label ?label_en . FILTER(LANG(?label_en) = "en") }
               }
               # remove possible ID duplicates
-              GROUP BY ?id}
+              GROUP BY ?id
+              ORDER BY ?id}
```
thomasneirynck: would enforcing this ordering have any unexpected effects? (just asking, probably not, but maybe missing something her)

sources/it/README.md
```diff
@@ -0,0 +1,51 @@
+# Italy Provinces
```
thomasneirynck: great write-up. thank you, crazy how complicated this gets.

templates/source_template.hjson
nickpeihl: I don't think so. I still do some post-processing on the data after retrieval to sort the features for better git diffing. So this isn't necessary here. I'll remove it.

PR #72, opened 2019-01-16
 Add South Korea Munisipalities

kimjmin:
This will add South Korean Munisipalities codes to Elastic Maps Service.

details at
sources/kr/README.md

cc: @nickpeihl


kimjmin:
@nickpeihl 
I guess did put wrong `id` type, which contains 17 digits.
Can you change `id` vaule? or we can roll this pull request back and I will update correct ID.

kimjmin:
Will close and send correct one.

PR #73, opened 2019-01-16
Add South Korea Munisipalities

kimjmin:
This will add South Korean Munisipalities codes to Elastic Maps Service.

details at
sources/kr/README.md


kimjmin:
@nickpeihl 
I don't get why build is failed. 
First commit was success and suddenly I understood that I didn't change `createdAt` and `id` from copied template file, so I changed that onto other commit. Then that commit is failed. 

Can you help me to figure out this? Can it be a timezone issue? probabily CI server's time is past than createdAt time?

PR #82, opened 2019-01-31
EMS v7.0

nickpeihl:
No big changes for EMS v7.0. Just need to make it available so we can add new v7 layers.


scripts/generate-manifest.js
```diff
@@ -83,6 +83,7 @@ function generateVectorManifest(sources, opts) {
           layers.push(manifestLayerV2(source, opts.hostname));
           break;
         case 6:
+        case 7: // v6 and v7 manifest schema are the same
```
thomasneirynck: oh man, the intricacies of cascading a switch/case statement :cold_sweat: ;) great idea of adding that comment

PR #100, opened 2019-04-30
Per README remove Cantone from label_it field in Switzerland cantons

nickpeihl:
This fixes a typo for one canton of Switzerland (Grigioni). Per the README, the data needs to be manually changed to remove "Canton of" equivalents from all translations of the `label_*` fields. 

Since this is a typo on a label field, I suggest we make the changes retroactive to all versions of EMS.


dschneiter:
@thomasneirynck what do you mean with  
> unlikely this will be used for joining.

To render Swiss data we should be able to use this in a join field. Cantons are similar to US States or French departments etc.

PR #104, opened 2019-05-07
[ci] add jjbb jobs and utility script

jonahbull:
Add the definitions for the current set of CI jobs so they can be
managed from this repo. The values in defaults.yml will be merged with
the job definitions. check_paths_for_matches is a utility script used to
prevent unnecessary builds from happening on every push to master due to
problems with the Jenkins git plugin.

cc: @spalger


nickpeihl:
lgtm

PR #111, opened 2019-07-02
Add ci job for feature-layers branch

nickpeihl:
To make new layer development easier to test, I propose developing new layers on the `feature-layers` branch. Any new layers being developed will be committed to this branch. This will build catalogue and vector manifests on a development storage bucket that are suitable for testing with Kibana by modifying the `map.manifestServiceUrl` parameter in `kibana.yml`. 

When ready to be reviewed and published, a pull request from the `feature-layers` branch will be opened against `master`. 

Code changes and pull requests not related to new layer development should continue to be created on new branches from `master`. The `feature-layers` branch would only be used for new vector layer development.

https://github.com/elastic/infra/issues/9313

### DIFF
--- a/deployDev.sh
+++ b/deployDev.sh
@@ -21,7 +21,6 @@ fi
 export EMS_PROJECT="emsfiles"
 
 export TILE_HOST="tiles.maps.elastic.co"
-export VECTOR_HOST="storage.googleapis.com/${GPROJECT}-${EMS_PROJECT}-vector-dev"
 export CATALOGUE_BUCKET=${GPROJECT}-${EMS_PROJECT}-catalogue-dev
 export VECTOR_BUCKET=${GPROJECT}-${EMS_PROJECT}-vector-dev
 export VECTOR_HOST="storage.googleapis.com/${VECTOR_BUCKET}"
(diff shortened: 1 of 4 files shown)


deployDev.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ -z "${GPROJECT}" ]]; then
+    echo "GPROJECT is not set, e.g. 'GPROJECT=elastic-bekitzur'"
+    exit 1
+fi
+
+export EMS_PROJECT="feature-layers"
+
+export TILE_HOST="tiles.maps.elastic.co"
+export VECTOR_HOST="storage.googleapis.com/${GPROJECT}-${EMS_PROJECT}-vector"
+export CATALOGUE_BUCKET=${GPROJECT}-${EMS_PROJECT}-catalogue
```
nyurik: i think you need a new set of buckets here - e.g. `-feature-catalogue` and `-feature-vector`

This also means you will need to create those two buckets, similar to https://github.com/elastic/infra/blob/master/terraform/providers/gcp/env/elastic-bekitzur/maps-file-service.tf#L92-L104

PR #123, opened 2019-09-04
Add Canton codes to Switzerland

### LINKED ISSUE #122: Add alpha 2 code for Switzerland Cantons
Selecting a data field containing the official 2-alpha canton codes does not seem to match the `ISO 3166-2 code` join field for the `Switzerland Cantons` vector map 

Screen Shot 2019-08-17 at 21 25 53

nickpeihl:
Fixes #122 

This PR adds a new field to the Switzerland cantons dataset containing the [two digit code](https://en.wikipedia.org/wiki/Data_codes_for_Switzerland#Cantons) for each canton. These codes are derived from but considered separate from the [full ISO 3166-2 code](https://www.iso.org/obp/ui/#iso:code:3166:CH). For example, the ISO 3166-2 code for Bern is "CH-BE". The two digit canton code is "BE".

Since this does not modify any existing fields and only adds a new field, I believe we can push this change to all versions of Elastic Maps Service without concern for breaking any users' existing visualizations.

Git diffs do not work well with large geojson files due to the way they are structured. My workaround is piping the outputs of the original and altered geojson files into jq and diff those results. 

For example:
`vimdiff <(cat data/switzerland_cantons_v1.geo.json | jq) <(curl --compressed https://vector.maps.elastic.co/files/switzerland_cantons_v1.geo.json?elastic_tile_service_tos=agree | jq)`

There are some differences in coordinate ordering in the geometry, but that should not make a difference for the end user.

@dschneiter I named the new field "Canton code" to avoid any confusion with the ISO 3166-2 codes that include the country code (e.g. CH-AG). But I am happy to discuss other alternatives.


dschneiter:
@nickpeihl In #122 I suggested `ISO 3166 Canton Code` as it directs the user to the proper specification (ISO 3166) and column.

I found that useful, because `ISO 3166 Canton Name` would have been another suggested *addition* as the localized names don't really make sense, as they are not localized, but mixed as in ISO 3166.

PR #126, opened 2019-09-18
Serbia and Montenegro

jsanz:
Added sources for Serbia and Montenegro.


sources/rs/serbia-districts.hjson
```diff
@@ -0,0 +1,142 @@
...
+            (if(bound(?id2),?id2,?id) as ?id)
+            ?iso_3166_2 ?label_en ?label_sr
+            WHERE {
+              SERVICE <https://query.wikidata.org/sparql> {
+              # Using nested query to ensure there is only one ?id2 value
+              {SELECT
+              ?id
+              (SAMPLE(?id2) as ?id2)
+              (SAMPLE(?iso_3166_2) as ?iso_3166_2)
+              (SAMPLE(REPLACE(?label_en," District","")) as ?label_en)
```
nickpeihl: :chef kiss: I like replacing the qualifiers in the SPARQL query. I wish it could be done for more countries.

jsanz: yeah, I realized it was doable directly on the query to avoid an extra step 😄

jsanz:
> Feel free to merge. Would you prefer using PRs for reviews rather than pushing upstream? I'm open either way.

I kind of prefer the PR process because we can discuss in a more comfortable place and it's easier to refer than specific commits.

Let me know if I should use any specific convention for tags and titles in order to skip or not automated stuff.

PR #130, opened 2019-10-17
Ukraine regions

jsanz:
Adding Ukraine regions (oblasts). On this country there is a current claim for Crimea, _de facto_ situation is that is controlled by Russia but according to [Wikipedia](https://en.wikipedia.org/wiki/Political_status_of_Crimea#Recognition_of_Ukraine's_territorial_integrity), the vast majority of UN recognizes it for Ukraine.

There is still a small mess around Sevastopol region.

image

### Wikidata

* Wikidata Ukraine Q212 entry points to Sevastopol [Q7525](https://www.wikidata.org/wiki/Q7525). This entry just refers to the city even it holds the `UA-40` ISO code, wikipedia entries, etc.

* There is also an almost empty item  [Q4456205](https://www.wikidata.org/wiki/Q4456205). It just contains a locator map, instance type, and nothing else apart from being referenced by [this Wikipedia article](https://en.wikipedia.org/wiki/Administrative_and_municipal_divisions_of_Sevastopol).

## OSM

* In the OSM side, the relation for the country [60199](https://www.openstreetmap.org/relation/60199) contains the region relation [1574364](https://www.openstreetmap.org/relation/1574364) as a subarea. That relation points to the almost empty [Q4456205](https://www.wikidata.org/wiki/Q4456205). This entry contains the correct ISO code.

* Relation [3030976](https://www.openstreetmap.org/relation/3030976) for the city boundaries pointing to [Q7525](https://www.wikidata.org/wiki/Q7525), hence producing the current geometry output.

---

@nyurik, what should we do in this case? I understand Ukraine (Q212) should have [Q4456205](https://www.wikidata.org/wiki/Q4456205) instead of [Q7525](https://www.wikidata.org/wiki/Q7525), as the first refers to the full region, but the second holds all the relevant info. The other option would be to modify OSM to make the region relation to point to the wikidata entry with all the details, ignoring the almost empty entry that seems to serve just to provide the locator map for the [Wikipedia article](https://en.wikipedia.org/wiki/Administrative_and_municipal_divisions_of_Sevastopol) :thinking:.


nickpeihl:
I propose adding the OSM relations to their respective Wikidata items as they are assigned in OSM. We could add the ISO code to [Q4456205](https://www.wikidata.org/wiki/Q4456205), ignoring (for now) the "distinct values constraint". Then we can swap the ids in the SPARQL query as we do with [Taiwan in China provinces](https://github.com/elastic/ems-file-service/blob/master/sources/cn/provinces.hjson#L73).

jsanz:
@nickpeihl thanks, as per your suggestion I added the OSM, ISO, and Country statements to [Q4456205](https://www.wikidata.org/wiki/Q4456205) and used the same technique to swap the entry for Sevastopol.

sources/ua/ukraine-regions.hjson
```diff
@@ -0,0 +1,137 @@
...
+      Ukraine Regions
+    ]
+    humanReadableName: {
+      ar: أوبلاستات أوكرانيا
+      ast: Oblast d'Ucraína
+      be: вобласць Украіны
+      be-tarask: вобласьць Украіны
+      bs: Oblasti u Ukrajini
+      ca: Oblast d'Ucraïna
+      cs: Oblast
```
nickpeihl: This label is too generic and does not mention the country. Unless you know the correct translation, I suggest removing it. 
```suggestion

```

sources/ua/ukraine-regions.hjson
```diff
@@ -0,0 +1,137 @@
...
+      fr: Oblast d'Ukraine
+      he: מחוזות אוקראינה
+      hr: Oblasti u Ukrajini
+      hsb: oblasć na Ukrainje
+      id: Oblast di Ukraina
+      it: Oblast' dell'Ucraina
+      ka: ოლქი
+      lv: Ukrainas apgabals
+      nb: Oblast i Ukraina
+      nl: Oblast
```
nickpeihl:
```suggestion

```

sources/ua/ukraine-regions.hjson
```diff
@@ -0,0 +1,137 @@
...
+      pl: Obwód na Ukrainie
+      pt: Oblasts da Ucrânia
+      pt-br: Oblasts da Ucrânia
+      ru: область Украины
+      sco: oblast o Ukraine
+      sh: Oblasti u Ukrajini
+      sk: Oblasť
```
nickpeihl:
```suggestion

```

sources/ua/ukraine-regions.hjson
```diff
@@ -0,0 +1,137 @@
...
+      es: Óblast de Ucrania
+      eu: Ukrainako oblast
+      fa: فهرست شهرهای اوکراین
+      fr: Oblast d'Ukraine
+      he: מחוזות אוקראינה
+      hr: Oblasti u Ukrajini
+      hsb: oblasć na Ukrainje
+      id: Oblast di Ukraina
+      it: Oblast' dell'Ucraina
+      ka: ოლქი
```
nickpeihl:
```suggestion

```

sources/ua/ukraine-regions.hjson
```diff
@@ -0,0 +1,137 @@
...
+            SELECT (if(bound(?id2),?id2,?id1) as ?id)
+                   ?iso_3166_2
+                   ?label_en ?label_uk ?label_ru
+            WHERE {
+              SERVICE <https://query.wikidata.org/sparql> {
+
+                SELECT ?id1
+                       (SAMPLE(?id2) as ?id2)
+                       (SAMPLE(?iso_3166_2) as ?iso_3166_2)
+                       (SAMPLE(?label_en) as ?label_en)
```
nickpeihl: Can we remove "область" and "Oblast" qualifiers from names in the geojson file, either by `(REPLACE...` in the SPARQL query or add a README with a note on manually removing them?

jsanz:
@nickpeihl not sure how the commits to Liechtenstein and Italy regions ended on this PR :disappointed:, let me know if you prefer for me to recreate it.

PR #132, opened 2019-10-22
Italy Regions

### LINKED ISSUE #91: Italy Regions
We should have a layer of the regions of Italy along with the existing provinces layer. According to several sources [[1](https://github.com/elastic/kibana/issues/33230#issue-420918516)] [[2](https://en.wikipedia.org/wiki/Provinces_of_Italy#Controversies)], regions are more commonly used than provinces in Italy. One region (Friuli-Venezia Giulia) has even [disbanded its provinces](http://www.consiglio.regione.fvg.it/cms/hp/informazioni/0072.html).

jsanz:
Closes #91 

The data was well and I didn't have to do anything special to get it using or regular SPARQL query.


sources/it/regions.hjson
```diff
@@ -0,0 +1,180 @@
...
+    }
+    {
+      type: property
+      name: label_en
+      desc: Province name (English)
+    }
+    {
+      type: property
+      name: label_it
+      desc: Province name (Italian)
```
nickpeihl: The `v2` manifest still uses the `desc` field for the field name. So let's change these accordingly. Versions >=`v6.6` use the localized field names.

```suggestion
      desc: Region name (Italian)
```

nickpeihl:
@markov00 can you look over the geojson file in this PR?

sources/it/regions.hjson
```diff
@@ -0,0 +1,180 @@
...
+  ]
+  fieldMapping: [
+    {
+      type: id
+      name: iso_3166_2
+      desc: ISO-3166-2 identifier
+    }
+    {
+      type: id
+      name: istat
```
nickpeihl: This should match the property in the geojson.

```suggestion
      name: istat_id
```

sources/it/regions.hjson
```diff
@@ -0,0 +1,180 @@
...
+    fur: Regjons de Italie
+    fy: Italjaanske Regio’s
+    gsw: Italienische Regionen
+    he: מחוזות איטליה
+    hr: Talijanske regije
+    hu: Olaszország régiói
+    ia: region de Italia
+    id: Daftar wilayah di Italia
+    is: Héruð Ítalíu
+    it: regione d'Italia
```
markov00: Where is used this field? Because `regione d'italia` is singular and is not the correct translation for `Italian regions`

nickpeihl: If or when Kibana is localized to Italian, this appears as the layer name in the select list. Is the correct plural form `regioni d'Italia`?

markov00: both `Regioni Italiane` or `Regioni d'Italia`  sounds fine for me

jsanz: I'll fix Spanish and Catalan for plurals and wait for @markov00 on Italian, thanks for your help!

markov00: Let's go with `Regioni d'Italia`
https://it.wikipedia.org/wiki/Regioni_d%27Italia

PR #143, opened 2019-11-20
Use relative URLs in v7.6 manifest

### LINKED ISSUE #141: [7.6] Urls in manifest should be relative
To make it easier for on-prem solutions, we should make the URLs in the 7.6 vector manifest relative to the current host. This will require additional changes to ems-client and possibly Kibana.

For example:

`https://vector.maps.elastic.co/v7.6/manifest`
```diff
{
  "layers": [
    {
      "layer_id": "world_countries",
      "created_at": "2017-04-26T17:12:15.978370",
      "attribution": [
        {
          "label": {
            "en": "Made with NaturalEarth"
          },
          "url": {
            "en": "http://www.naturalearthdata.com/about/terms-of-use"
          }
        },
        {
          "label": {
            "en": "Elastic Maps Service"
          },
          "url": {
            "en": "https://www.elastic.co/elastic-maps-service"
          }
        }
      ],
      "formats": [
        {
          "type": "geojson",
-          "url": "https://vector.maps.elastic.co/files/world_countries_v1.geo.json?elastic_tile_service_tos=agree",
+          "url": "files/world_countries_v1.geo.json",
          "legacy_default": true
        }
      ],
...
```

nickpeihl:
Fixes #141. To make it easier to reverse proxy and deploy locally, we are changing the URLs in the v7.6 manifest from absolute to relative. Additionally, the `?elastic_tile_service_tos=agree` query string is no longer hard-coded in the manfiest. Clients using the v7.6 API will need to accept the [EMS Terms of Service](https://www.elastic.co/elastic-maps-service-terms) by appending the querystring `?elastic_tile_service_tos=agree` to all requests.


thomasneirynck:
Not sure if I have this 100% right, so bear with me ;)

The relative-path is from the root. 

If the client is configured to the manifest (which is `host/version/manifest`), the client would need knowledge about this relationship. So this would complicate if users put EMS behind a proxy like so:

`reverseproxy.org:5401/tiles/*` -> `tiles.maps.elastic.co/v7.6`
`reverseproxy.org:5401/files/*` -> `vector.maps.elastic.co/v7.6`

Unless we somehow bake this knowledge into Kibana/user-docs that the sub-path of the manifest is _not_ the top-level (and imho, I think we actively want to avoid such a situation since this is the situation it is _today_).

We could update the EMS-client it needs a double config:

```
emsFileConfig:
     host: reverseproxy.org:5401/tile
     version: v7.2
```

With the understanding on the back-end that EMS _always_ needs to serve the manifest from `host/version/manifest`. Then the actual files have their relationship from `host/relative-url-from-manifest`.
 
Note that EMS diverges significantly from other geo-server products (e.g. ArcGIS Server, GeoServer, ...). Most of these platforms generate their manifests dynamically, based on user-config of the host. Mapbox-tile-spec are an example of this as well. 

Another approach would be more along the lines of organizing resources hierarchically. (e.g. https://sampleserver6.arcgisonline.com/arcgis/rest/services/). The above is counter to a how EMS was designed (efficient storage of files, no duplication etc...).

Thoughts?

nickpeihl:
> We could update the EMS-client it needs a double config:
> 
> ```
> emsFileConfig:
>      host: reverseproxy.org:5401/file
>      version: v7.6
> ```

ems-client already accepts a `kbnVersion` parameter that gets passed as a query parameter in requests for tiles and files. We could [parse this parameter](https://github.com/elastic/ems-client/pull/13/commits/88acfcb7215c6b60ec691f21f27f7ee6d838fa9a) to get the matching EMS version. Although, this requires us to publish the next [major](https://github.com/elastic/kibana/blob/master/package.json#L14) and [minor](https://github.com/elastic/kibana/blob/7.x/package.json#L14) releases of EMS for development. However, these future EMS releases would be subject to change until stack release. 

> Another approach would be more along the lines of organizing resources hierarchically. (e.g. https://sampleserver6.arcgisonline.com/arcgis/rest/services/). The above is counter to how EMS was designed (no duplication of files for efficient content etc...).
> 

I would prefer to keep the current convention to avoid duplicating vector files. Especially as we move to [syncing EMS releases with the stack](https://github.com/elastic/ems-file-service/issues/144) and also if we start to release higher resolution (much larger) datasets.

nickpeihl:
> ems-client already accepts a `kbnVersion` parameter that gets passed as a query parameter in requests for tiles and files. We could [parse this parameter](https://github.com/elastic/ems-client/pull/13/commits/88acfcb7215c6b60ec691f21f27f7ee6d838fa9a) to get the matching EMS version. Although, this requires us to publish the next [major](https://github.com/elastic/kibana/blob/master/package.json#L14) and [minor](https://github.com/elastic/kibana/blob/7.x/package.json#L14) releases of EMS for development. However, these future EMS releases would be subject to change until stack release.

One concern I have with this is handling breaking changes in future releases and functional tests downstream. For example, we might make a change to the v8.0 manifest while still in development that could require functional test changes in Kibana.  The CI on the `master` branch and other PRs would fail until the functional test changes are fixed. We may be able to minimize the effect with parallel PRs, but I do not see how we could avoid it completely.

test/manifest-v7.6.js
```diff
@@ -0,0 +1,205 @@
...
+  t.false(catalogue, '7.6 catalogue should not exist');
+
+  // It is not necessary to test different hostnames, since URLs in manifest are relative
+  const vector = generateVectorManifest(sources, {
+    version: 'v7.6',
+    hostname: 'vector.maps.elastic.co',
+    fieldInfo: fieldInfo,
+  });
+  t.deepEquals(vector, vectorExpected, 'v7.6 vector manifest');
+  t.end();
```
jsanz: Not sure if this is needed or helpful, but maybe we could check here not just `7.6` but future `v7` releases and that it fails for `v8` (so we can ensure we control that correctly later this year).

```js
tap('>=7.6 tests', t => {

  ['v7.6', 'v7.8', 'v7.8.1'].forEach(version => {
    const catalogue = generateCatalogueManifest({
      version: version,
      tileHostname: 'tiles.maps.elstc.co',
      vectorHostname: 'vector.maps.elastic.co',
    });
    t.false(catalogue, version + ' catalogue should not exist');

    // It is not necessary to test different hostnames, since URLs in manifest are relative
    const vector = generateVectorManifest(sources, {
      version: version,
      hostname: 'vector.maps.elastic.co',
      fieldInfo: fieldInfo,
    });
    t.deepEquals(vector, vectorExpected, version + ' vector manifest');
  });

  //v8 vector manifest should fail at this moment
  tap('Check that v8.0 fails', function (t) {
    t.throws(function () {
      generateVectorManifest(sources, {
        version: 'v8.0',
        hostname: 'vector.maps.elastic.co',
        fieldInfo: fieldInfo,
      });
    }, new Error('Unable to get a manifest for version 8.0.0'), 'throws assert error');
  });

  t.end();
});
```

Issue #53, opened 2018-09-18
Test for existence of join fields in vector data


nickpeihl:
We should have unit tests that verify that all join fields in the manifest files actually exist in the vector data. This would allow us to catch data errors like in #51 prior to publishing.

jsanz:
After discussion at #152 @nickpeihl and I agreed that it's OK to have fields not present in the data. This is the case for labels in a few countries where some regions don't use a language used in other parts of the country (reminder that those labels are marked as `OPTIONAL` in the SPARQL query).

Still, an optional test has been added and documented on the repo `README.md` to check this condition.

PR #146, opened 2019-11-26
Spain provinces and autonomous communities

### LINKED ISSUE #142: Split Spain boundaries into provinces and regions
At this moment Spanish boundaries store together both provinces and autonomous communities, being the second a higher admin level. We should take out the autonomous communities from the current province's source. 

Since all boundaries have an ISO code we could not do this, since joining with other data will yield the correct data, but for consistency and clarity, like in other countries like the USA or Italy, it's better to have them in different sources.

jsanz:
fixes #142 

This PR splits Spanish provinces and autonomous communities (something similar to USA counties and states) into two separated layers.

As described in #142, a few wikidata entries had two different ISO codes, one for the province and another for the autonomous community.

To retrieve the data for the provinces, since relations in OSM were properly modeled (after adding a few Wikidata tags), the SPARQL query is quite different from the typical ones we've been using.


jsanz:
@nickpeihl please take a look when you have some time and let me know what you think

jsanz:
@nickpeihl thanks for your comments :raised_hands: 

I've moved the id's to the feature level using the script below. Should I put it in the scripts folder?

Also, should this check be included somehow in the test suite? I don't know exactly how we check geometries on the test process, but it would be great if we could ensure the ids are correctly placed.

<details>
<summary>Script to move id's</summary>

```js
/*
 * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
 * or more contributor license agreements. Licensed under the Elastic License;
 * you may not use this file except in compliance with the Elastic License.
 */

const fs = require('fs');

try {
  const filePath = process.argv[2];

  if (!filePath) {
    throw new Error(`Add the GeoJSON file path to process. e.g. node scripts/move-ids.js data/usa_states_v1.geo.json`);
  }

  const file = fs.readFileSync(filePath, 'utf8');
  const geojson = JSON.parse(file);

  if (geojson.features[0].id) {
    throw new Error('File already has ID\'s at feature level, nothing to do');
  }

  // overwrite the features moving the ID field at the feature level
  geojson.features = geojson.features.map(({ type, geometry, properties }) => {
    const { id, ...propsWithoutId } = properties;
    return { id, type, geometry, properties: propsWithoutId };
  })

  process.stdout.write(JSON.stringify(geojson))
} catch (error) {
  console.log(error.message)
  process.exit(1);
}
```
</details>

nickpeihl:
> Also, should this check be included somehow in the test suite? I don't know exactly how we check geometries on the test process, but it would be great if we could ensure the ids are correctly placed.

I looked at our existing data and having a top-level `id` property for each feature is not a hard constraint. For example, South Korea municipalities do not have top level feature ids because it comes from a different data source. Most of our other data is generated from SPARQL which automatically creates the id property.

So I am going to backtrack a bit here. Sorry, but now I __do not think__ we need to require every feature have a top-level `id` property. But maybe we should constrain feature properties to only those specified in the source metadata (essentially the reverse of [this issue](https://github.com/elastic/ems-file-service/issues/53))?

This PR lgtm! Let's merge this and consider the other questions separately.

jsanz:
> So I am going to backtrack a bit here. Sorry, but now I do not think we need to require every feature have a top-level id property. But maybe we should constrain feature properties to only those specified in the source metadata (essentially the reverse of this issue)?

Shall we open a new issue to track this? This should be placed in the testing logic right?

nickpeihl:
Yes, this is would be in the testing logic. I have created [an issue](https://github.com/elastic/ems-file-service/issues/151) to track this. Thanks.

PR #147, opened 2019-11-26
Docker image

### LINKED ISSUE #145: [7.6] Docker image
We should have a Docker image for hosting EMS vector manifests and files starting with the 7.6 release. The Docker build process should generate and serve the `/dist` as a web service.

jsanz:
This is a minimal addition to the project to support generate a Docker image that exposes EMS as a static web service.

The Dockerfile expects the project to be built and the assets ready at the `dist` folder.

A simple configuration is provided, it serves the folder contents for easy testing now, but we may want to change that one, and maybe other settings, before releasing.

Proposing this to be merged into `master` since it does not affect any work we are doing on the data side at this moment, even it is related with #143 

fixes #145

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.gitignore
+node_modules
+.editorconfig
+.eslint*
+README
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,32 @@ FROM nginx:alpine
 
 LABEL MAINTAINER "Jorge Sanz <jorge.sanz@elastic.co>"
 
-ARG ARG_VECTOR_HOST="localhost"
+# Bring environment variables and default values
+ARG ARG_HTTP_PROTOCOL="https://"
+ENV HTTP_PROTOCOL=${ARG_HTTP_PROTOCOL}
+
+ARG ARG_VECTOR_HOST="files.maps.elastic.co"
 ENV VECTOR_HOST=${ARG_VECTOR_HOST}
+
+ARG ARG_TILE_HOST="tiles.maps.elastic.co"
+ENV TILE_HOST=${ARG_TILE_HOST}
+
+ARG ARG_HTTP_PORT="80"
+ENV HTTP_PORT=${ARG_HTTP_PORT}
+
+ARG ARG_VECTOR_PATH=""
+ENV VECTOR_PATH=${ARG_VECTOR_PATH}
+
+ARG ARG_TILE_PATH=""
+ENV TILE_PATH=${ARG_TILE_PATH}
+
+ARG ARG_DOCKER_IMAGE_NAME="ems-file-service"
+ENV DOCKER_IMAGE_NAME=${ARG_DOCKER_IMAGE_NAME}
+
+ARG ARG_DOCKER_IMAGE_VERSION="0.0.1"
+ENV DOCKER_IMAGE_VERSION=${ARG_DOCKER_IMAGE_VERSION}
+
+EXPOSE ${HTTP_PORT}
 
 # Copy sources
 RUN rm -rf /usr/share/nginx/html/*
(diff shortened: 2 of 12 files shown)


jsanz:
@nickpeihl I'm asuming here that the `dist` folder is generated but since I don't know the internals on how an image would be generated please let me know if I should also generate the full thing asuming a clean checkout from the repository. I can install `nodejs`, build the assets, and then remove the software, if that makes the process easier to integrate.

jsanz:
OK @nickpeihl so `VECTOR_HOST` is a configuration for Nginx, but `TILE_HOST` is a variable to change in the manifests at running time, not when building the image, right?

Umm, maybe there's something we can do using Nginx to replace strings when serving static data, but otherwise, we will have to put some dynamic language in between, since the manifests are built when the image is generated, not when the container starts.

I'll continue working on this :muscle:

jsanz:
I will try to build everything using Docker `ENTRYPOINT` command to generate the manifests when the image starts :+1:

jsanz:
Did quite a few changes to building code to:

* use current `build.sh` script to also generate a local Docker image
* created a `deployDockerImage.sh` with some default environment variables that allow to set up custom hostnames, protocol and path in case files are served from a single domain for both catalogue and EMS files
* added support for these custom values in the manifests generation
* added testing for catalogue and vectors (we may want to add more test cases here)

To try this:

1. Run `./deployDockerImage.sh` to generate the local image
2. Run the image with `docker run --rm --name ems -p 8000:80 ems-file-service:0.0.1`
3. Add this line to your `/etc/hosts` file: `127.0.1.1       custom.files.maps.elastic.co`
4. Try to access `http://custom.files.maps.elastic.co:8000/` and check if both cataloge and vector files folders show up

You can check if the full chain of calls is working running:

```sh
$ curl -s custom.files.maps.elastic.co:8000/catalogue/v7.2/manifest \
	| jq -r ".services[1].manifest" \
	| xargs curl -s \
	| jq -r ".layers[0].formats[0].url" \
	| xargs curl -I
```

That should return a `200` response from the server meaning that:

* Catalogue exists and points to the files service
* The files service catalog exists, has layers, and the url for the first layer exists as well

jsanz:
additionally to the last update I checked that the Docker image can work also as a proxy for the EMS basemaps server using a configuration like this one:

```
  location ~ /${TILE_PATH}(.*) {
    resolver 8.8.8.8;
    gzip off;

    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-NginX-Proxy true;
    proxy_set_header Host $http_host;
    proxy_set_header Accept-Encoding "";

    proxy_pass http://tiles.maps.elastic.co/$1;

    proxy_ssl_session_reuse off;
    proxy_cache_bypass $http_upgrade;
    proxy_redirect off;

    sub_filter 'https://tiles.maps.elastic.co/' '${HTTP_PROTOCOL}${TILE_HOST}:${HTTP_PORT}/${TILE_PATH}';
    sub_filter 'http://tiles.maps.elastic.co/' '${HTTP_PROTOCOL}${TILE_HOST}:${HTTP_PORT}/${TILE_PATH}';

    sub_filter_once off;
    sub_filter_types '*';
  }
```

This configures a reverse proxy in the same instance that forwards to `tiles.maps.elastic.co` any request sent to `${TILE_PATH}` (assuming that both `VECTOR_HOST` and `TILE_HOST` are the same). The `sub_filter` directives replace from the proxied content hardcoded references to the EMS Basemaps server to the parameters set when building the image. This worked well in my environment except for some fonts that still were retrieved from `tiles.maps.elastic.co`.

Resources requested to the sample `myems.com`, both EMS vector files and proxied requests to basemaps resources (styles, sprite, vector tiles):
image


Fonts still requested to Elastic production server: 
image (1)

jsanz:
I'm going to close this PR since the work done here is not relevant anymore.

PR #149, opened 2019-11-27
Fix France department IDs

jsanz:
Related to the issue raised at #146, generated by the `fixgeometries.py` script that relocates the `id` property from the feature to the `properties` object, this layer was also affected.


jsanz:
Not much to check @nickpeihl, but I preferred to fix this layer in a separate PR, I hope this is OK.

PR #153, opened 2019-12-17
Feature layers Iceland Greenland

jsanz:
Adding boundaries for Greenland municipalities and Iceland regions.


jsanz:
@nickpeihl I this can be the last new layer to add to `feature-layers` before creating a new PR to `master`. 

This work I had to do small changes in Wikidata to update changes for Greenland municipalities that happened in 2018.

For Iceland, I took the path to retrieve the boundaries from their category [`Q842100`](https://www.wikidata.org/wiki/Q842100), instead of starting from the country, since the categorization at the country level seems to be wrong (I'm looking into this just to see if I can improve Wikidata entry, without blocking this work).

jsanz:
OK yep, indeed the definition is different, let's merge this and we can always improve OSM data in the future if it makes sense.

PR #157, opened 2020-01-09
Recovered fixed geometries for France

jsanz:
The correct geometries for France departments

image


jsanz:
this was the layer with very [odd geometries](#137) that ended up needing QGIS to fix them, so probably that's the reason for having multis.

thanks!

PR #158, opened 2020-01-13
Fix Bremen state of Germany name

jsanz:
fixes https://github.com/elastic/kibana/issues/54520

Update the SPARQL query to remove the string `Freie Hansestadt ` from states names.


jsanz:
That was a change made in Wikidata in [June 2018](https://www.wikidata.org/w/index.php?diff=708698563&oldid=700902926&title=Q1196&type=revision
). On the other hand [Wikipedia](https://en.wikipedia.org/wiki/Mecklenburg-Vorpommern) and the state [Official website in English](https://en.mecklenburg-vorpommern.de/homepage/) both use the German form so I think the anglicized version is not accepted anymore. ISO website is not showing any other official form than in German as well.

PR #160, opened 2020-01-15
Merge pull request #160 from jsanz/layerdiff

jsanz:
A bit tired of manually checking differences between versions of data layers, I've created this script that takes two GeoJSON files and compares in a few different ways, trying to catch meaningful insights.

The default thresholds may need a bit of tuning as we use the script in future data tasks, but so far the tool looks it should make the job.

### DIFF
--- a/scripts/layerdiff.js
+++ b/scripts/layerdiff.js
@@ -1,0 +1,231 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+const fs = require('fs');
+const _ = require('lodash');
+const jsts = require('jsts');
+const distance = require('turf-vincenty-inverse');
+
+/* Command line options definition */
+const yargs = require('yargs')
+  .version()
+  .scriptName('layerdiff')
+  .help()
+  .alias('h', 'help')
+  .option('verbose', {
+    alias: 'v',
+    default: 0,
+    count: true,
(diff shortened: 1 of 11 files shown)


scripts/layerdiff.js
```diff
@@ -0,0 +1,335 @@
...
+
+  const leftPath = argv._[0];
+  const rightPath = argv._[1];
+
+  const reader = new jsts.io.GeoJSONReader();
+
+  // Load GeoJSON files
+  print(`================================== Loading files...`);
+  const leftJson = reader.read(fs.readFileSync(leftPath, 'utf-8'));
+  const rightJson = reader.read(fs.readFileSync(rightPath, 'utf-8'));
```
nickpeihl: I would also find it useful to be able to pass in a GeoJSON string from stdin, maybe by using `-` as a parameter. Then I could do something like `git show c0b30f:data/france_departments_v7.geo.json | node scripts/layerdiff.js - data/france_departments_v7.geo.json ` to compare a previous revision or different branch to the file on the working branch. 

Something like this, maybe? What do you think @jsanz?
```suggestion
  const leftJson = leftPath === '-'
    ? reader.read(fs.readFileSync('/dev/stdin', 'utf-8'))
    : reader.read(fs.readFileSync(leftPath, 'utf-8'));
  const rightJson = rightPath === '-'
    ? reader.read(fs.readFileSync('/dev/stdin', 'utf-8'))
    : reader.read(fs.readFileSync(rightPath, 'utf-8'));
```

jsanz: thanks! I had that idea but was not sure how to implement it, I changed `/dev/stdin/' by `0` which apparently make it more portable. Change in 21925f1

PR #161, opened 2020-01-15
Add EMS version to manifests

### LINKED ISSUE #159: Add version number to manifest
Manifests are versioned, but we don't publish them with the actual version number. this is now implicit in the URL where it is hosted. Ideally, we publish the version number inside the manifest.

A similar ask for the tile-service.

nickpeihl:
Partially fixes #159 

This PR adds a root level `version` property to all catalogue and vector file manifests containing the EMS version in the form of "{major}.{minor}" (e.g. 1.0, 2.0, 6.6, 7.2). 

Since this only adds a new property and does not modify the schema of the manifests otherwise, I believe this is safe to apply retroactively to all EMS versions. I have successfully tested these manifests in Kibana 5.6.14, 6.8.0, 7.1.1, 7.2.1, and the 7.6 branch.

### DIFF
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -39,6 +39,7 @@ function generateCatalogueManifest(opts) {
     ? { id: 'tiles_v2', version: 'v2' }
     : { id: 'tiles', version: 'v7.2' };
   const manifest = {
+    version: `${version.major}.${version.minor}`,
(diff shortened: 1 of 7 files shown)


jsanz:
I tested on `master` but I was not able to test this on previous versions. The code changes look good to me.

PR #167, opened 2020-03-31
Bump dependencies and remove package-lock.json

nickpeihl:
Remove package-lock.json and recommend yarn instead of npm as that's the precedent in use in Kibana.

Also, this bumps many dependencies.


nickpeihl:
> I had to upgrade Node to `12` to install packages because `jsts` complained, should we add a `{ "engines" : { "node" : ">=12" } }` to make this explicit on the `package.json`?

Good idea. I added this. The `build.sh` script uses the `node:lts` docker image which is currently Node 12, so I don't think we need to change that. It would fail CI anyways if the Docker image was not compatible.

PR #178, opened 2020-07-08
Administrative divisions layer

### LINKED ISSUE #177: Natural Earth Provinces and States layer
This issue tracks the work to improve [Natural Earth Admin 1](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/) dataset. @nickpeihl has already reported the regions that need some fix on [a google spreadsheet](https://docs.google.com/spreadsheets/d/1a9_Ib7Lha1l5yqHxPssUX1tWVziegQXws8No-006PC4/edit?usp=sharing).

@nickpeihl and I will be fixing geometries and eventually merge all together in a new dataset that would be merged with original NE data to produce the dataset to be served on EMS.

nickpeihl:
Fixes #177. 

# Administrative divisions

The Administrative divisions layer contains second level subdivisions (first level where no second level subdivision exists) of world countries. This layer was derived from the [Admin 1 - States, Provinces](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/) layer from [Natural Earth](https://www.naturalearthdata.com) with supplemental boundaries from [OpenStreetMap](https://www.openstreetmap.org) where Natural Earth data is known to be incomplete or erroneous.

When released, this layer will be available immediately in all releases of Kibana from v5.6+. The layer is published as a GeoJSON file for Kibana <v6.2. Starting with Kibana v6.2, a slightly higher resolution TopoJSON file is used. 

This layer can be viewed at https://maps.elastic.co/?manifest=testing#file/administrative_divisions. 

## Test in Kibana

[Instructions here](https://github.com/elastic/ems-file-service/pull/178#issuecomment-655219909).

## Notes
The ISO 3166-2 codes in this layer have been verified against the MaxMind GeoLite2 database currently distributed with the latest releases of Elasticsearch. Some features in the Natural Earth dataset this layer is derived from use unofficial ISO codes or have no ISO codes at all. 

This layer only offers one opinionated view of country boundaries. Generally, these boundaries are drawn according to [de facto](https://en.wikipedia.org/wiki/De_facto) rather than [de jure](https://en.wikipedia.org/wiki/De_jure) status.

## License

Since this layer contains data from both Natural Earth ([Public Domain](https://www.naturalearthdata.com/about/terms-of-use/)) and OpenStreetMap ([ODbL license](https://www.openstreetmap.org/copyright)) the more restrictive license (ODbL) applies to this dataset. 

Any use of Elastic Maps Service data and APIs are subject to the [Elastic Maps Service Terms of Service](https://www.elastic.co/elastic-maps-service-terms).


thomasneirynck:
you can test in Kibana Maps by adding this to the `kibana.dev.yml`.

```
map.emsFileApiUrl: https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev/
```

Packetbeat is probably the easiest way to load data. Surf to some global websites and see them pop up.

sources/world/admin_divisions_v2.hjson
```diff
@@ -0,0 +1,177 @@
...
+      {
+        label: {
+          en: Elastic Maps Service
+        }
+        url: {
+          en: https://www.elastic.co/elastic-maps-service
+        }
+      }
+    ]
+    fieldMapping: [
```
jsanz, line 34: Should we aim to add also the wikidata identifier? I know there are quite a few divisions that miss that field (almost 500) and I'm not sure how much that field is leveraged by our users.

jsanz:
Following @thomasneirynck suggestion on using the staging URL I got it working correctly so my previous comment was related with something on my environment.

Tested it using the Kibana Sample flights dataset and joining by the `DestRegion` field. :ok_hand: 

image

sources/world/admin_divisions_v1.hjson
```diff
@@ -0,0 +1,171 @@
...
+        label: {
+          en: Elastic Maps Service
+        }
+        url: {
+          en: https://www.elastic.co/elastic-maps-service
+        }
+      }
+    ]
+    fieldMapping: [
+        {
```
thomasneirynck, line 35: The field-names/descriptions are kind of hard to wrap your head around when you see them displayed:

e.g.: The field-descriptions looks like in the Maps-UX

image


If we ingest this file into Kibana Maps (e.g. using GeoJson upload), and use the raw documents, we see the field-names.

image

I'd consider renaming and using the ECS names for these fields.

```
 {
          type: id
          name: region_iso_code
          desc: iso2 region code
}
{
          type: id
          name: country_iso_code
          desc: iso2 country code
}
```


[EDIT] nevermind, I see that we already made some changes. Just wondering if we should be using the ECS-names.

sources/world/admin_divisions_v1.hjson
```diff
@@ -0,0 +1,171 @@
...
+          type: id
+          name: iso_3166_2
+          desc: ISO-3166-2 identifier
+        }
+        {
+          type: id
+          name: iso2
+          desc: Country code
```
thomasneirynck, line 43: Country code is the first id-field we're introducing in EMS that is _not_ a unique identifier.

The joining-code in Kibana Maps works fine enough. 

e.g. join on country-code. It captures all the individual regions.

image


image

The logic also works for region maps.

image

There's some edge-case, e.g. the display shows internal borders.


I guess the overall question is. Do we want to make this a new semantic of the id-field? ie. a field that uniquely identifies some entity, but it can be both a primary key (ie. the region-code) _or_ a foreign key (ie. the country-code).

thomasneirynck: ^ this behavior of the inner-join code is not unit-tested in Kibana (all tests expect a primary key in the left-hand side "table").

sources/world/admin_divisions_v1.hjson
```diff
@@ -0,0 +1,171 @@
...
+        }
+        {
+          type: property
+          name: name
+          desc: Division name
+        }
+    ]
+    name: 'administrative_divisions'
```
thomasneirynck, line 51: consider renaming to `Administrative regions`, or `2nd level regions`, ... Something to tie it to ECS more closely.

sources/world/admin_divisions_v2.hjson
```diff
@@ -0,0 +1,177 @@
...
+      zh-hant: 行政領土實體
+      zh-hk: 行政領土實體
+      zh-mo: 行政領土實體
+      zh-my: 行政领土实体
+      zh-sg: 行政领土实体
+      zh-tw: 行政領土實體
+    }
+    emsFormats: [{
+      type: geojson
+      file: 'admin_divisions_v2.geo.json'
```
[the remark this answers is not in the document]
thomasneirynck: Yes, it's not used. The landing page shows "Download TopoJson" only and Kibana uses default format (TopoJson).

I actually tested the GeoJson-version, and it works fine in the GeoJson upload of Kibana Maps.

As a tangent: we should consider improving the landing-page so the "Download GeoJson" button is always "Download GeoJson" if there's a geojson-version available. Right now, it defaults to default format (in this case, topojson). Topojson isn't the greatest interchange format anyway, GeoJson is more shareable/hackable.

jsanz:
Update on the issue loading this layer. I could reproduce the issue just enabling the proxy. Running Kibana from `7.8`, `7.7`, and `7.6` branches with these settings gives the same error I shared

```yaml
map.proxyElasticMapsServiceInMaps: true
map.emsFileApiUrl: https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev
```

Peek 2020-07-08 15-40

No traces in the server or browser logs. This indeed looks like some malfunction in the proxy.

sources/world/admin_divisions_v1.hjson
jsanz: I also prefer something that is close to the wording we see in ECS and the results of the geoip processor. +1 for `Administrative regions`

sources/world/admin_divisions_v2.hjson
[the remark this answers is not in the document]
jsanz: Agree that TopoJSON is not a very portable format.

Maybe an improvement could be to always add a GeoJSON equivalent file in the service and then favor it on the landing page as you suggest. In production, we only use TopoJSON at this moment for USA counties and zip codes.

sources/world/admin_divisions_v2.hjson
thomasneirynck: I think related to the overall _id_ question. Right now, I am not sure e.g. how useful it is to have the country-code in here. Will it ever be used for joining? if not, is there a longer term analytical purpose? (maybe towards hierarchical layers or something...) 

How common is the wikidata identifier, is this something our user-base will have in their own business-data? 

My gut feeling right now, is I think it's probably more useful to add the wikidata identifier than the country-identifier. (providing the wikidata-id is somewhat useful, and not just some obscure internal id with little outside use).

sources/world/admin_divisions_v2.hjson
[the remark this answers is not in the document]
nickpeihl: Yes, my intention was to include the geojson for future use such as ingest into Kibana by downloading from the landing page.

sources/world/admin_divisions_v2.hjson
nickpeihl: > I think related to the overall _id_ question. Right now, I am not sure e.g. how useful it is to have the country-code in here. Will it ever be used for joining? if not, is there a longer term analytical purpose? (maybe towards hierarchical layers or something...)

I think it may be useful to have the country code when the data layer is ingested into Elasticsearch so that it can be used in filters. But maybe it doesn't make sense to expose the field in the join.

> How common is the wikidata identifier, is this something our user-base will have in their own business-data?

It may not be very useful for joining to business data. Additionally, some divisions do not have a wikidata ID because I did not enforce it in my changes to the Natural Earth layer. 😬

thomasneirynck: my 2 cents, (and don't feel strongly about either, so for your consideration mostly).

- would _not_ include the wikidata-id. Just more content to maintain with questionable usefulness.
- would _not_ include the country-identifier. While it's true that it can be used for filtering, this can only be done if the data is ingested in Elasticsearch. This is not the most common use-case. There are work-arounds too: user can always write a manual filter (e.g. `region_code: US-*` or `region_code: CA-*` to filter about country. Having it included would complicate the field-types of EMS. It's not a unique identifier (ie, "type=id"), and neither is it useful for human readable labels ("type=label").

In either case, we can always add a new column later, without affecting users of the layers.

nickpeihl: Sounds good to me. I've only included the region iso code and region name in the latest update.

sources/world/admin_divisions_v1.hjson
nickpeihl: Field names and descriptions now match the ECS names.

sources/world/admin_divisions_v1.hjson
nickpeihl: I don't want to make broad changes here. I've removed the country code in the latest update.

nickpeihl:
> Update on the issue loading this layer. I could reproduce the issue just enabling the proxy. Running Kibana from `7.8`, `7.7`, and `7.6` branches with these settings gives the same error I shared
> 
> ```yaml
> map.proxyElasticMapsServiceInMaps: true
> map.emsFileApiUrl: https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev
> ```
> 
> Peek 2020-07-08 15-40
> 
> No traces in the server or browser logs. This indeed looks like some malfunction in the proxy.

It definitely looks like the proxy does not like having multiple formats. Even though the topojson format is the default in the manifest, the proxy returning the geojson format. I'm still trying to determine if that's a bug in the ems-client `fileLayer.getDefaultFormat` method or somewhere else in the proxy server.

nickpeihl:
With https://github.com/elastic/ems-file-service/pull/178/commits/c6abbed8d4881f8c30fc6f73246ac7dc243f0362 I am no longer seeing the error when using the layer via proxy. @jsanz can you confirm?

There appears to be a bug in Kibana where the proxied manifest does not work with multiple formats.

jsanz:
>  @jsanz can you confirm?

I can confirm, the layer manifest now reflects the correct format for the dataset and Kibana Maps loads it without issues.

Thanks for reporting the bug also! :muscle:

nyurik:
Per skype discussion, I think we should include disputed regions here as overlapping shapes.  Topojson makes this very efficient, essentially storing each shape just once.

For example, `TW` (Taiwan -- country level as having no subdivisions) and `CN-TW` [Taiwan Province](https://en.wikipedia.org/wiki/Taiwan_Province,_People%27s_Republic_of_China) would both be included as the same (or at least overlapping) shape with different metadata.

This will ensure that user's data really drives the (choropleth) visualization -- using whichever codes it actually contains. Note that if we pick just one shape, we risk user's data not being shown at all.  I think it is unlikely for the user's data to contain both IDs for the same region, and even if it does, it is better to show a strange coloration than no coloration at all (or an error of a missing region).

nickpeihl:
> Per skype discussion, I think we should include disputed regions here as overlapping shapes. Topojson makes this very efficient, essentially storing each shape just once.
> 
> For example, `TW` (Taiwan -- country level as having no subdivisions) and `CN-TW` [Taiwan Province](https://en.wikipedia.org/wiki/Taiwan_Province,_People%27s_Republic_of_China) would both be included as the same (or at least overlapping) shape with different metadata.
> 
> This will ensure that user's data really drives the (choropleth) visualization -- using whichever codes it actually contains. Note that if we pick just one shape, we risk user's data not being shown at all. I think it is unlikely for the user's data to contain both IDs for the same region, and even if it does, it is better to show a strange coloration than no coloration at all (or an error of a missing region).

Thanks Yuri. Looks like the provinces of Macau and Hong Kong are also missing. I'm working on those fixes.

nickpeihl:
Merging this now. But we will release this to production at a later date.

PR #184, opened 2020-09-02
Fix Bosnia and Herzegovina regions

### LINKED ISSUE #183: Review Bosnia-Herzegovina regions
Bosnia-Herzegovina region `BAH-BIH` appears repeated several times in the new regions dataset

image

We need to check if those are all part of a single bigger region and also review the rest of the country features.

cc. @nickpeihl

nickpeihl:
Fix #183 

This fixes the regions of Bosnia and Herzegovina. This PR also includes minor variations in other geometries. This may have been caused by using a different version of mapshaper.


jsanz (approved):
:+1: thanks for the last fixes!

PR #188, opened 2020-09-21
World Countries update

### LINKED ISSUE #164: Add population data to EMS files
Reading over https://www.esri.com/arcgis-blog/products/product/mapping/mapping-coronavirus-responsibly/ and one point for accurate choropleth maps is to normalize values with population data so values can be accurately compared across regions.

The first step to allowing this in Kibana would be to include population data in EMS. Then, Kibana could allow for a math expression in the join metric to do the normalization.

### LINKED ISSUE #179: Update World Countries Layer
The geometry in the [world countries layer](https://maps.elastic.co/#file/world_countries) is very low resolution and often jagged. We could update this layer to a moderately higher resolution that is more consistent with the upcoming administrative regions layer. 

I propose we do an in place update for all versions of EMS. 

The updated geojson layer may be around 5MB which is considerably larger than the current 627KB version. But we can experiment with simplification to get it smaller if necessary. Alternatively, we can limit the updated layer to EMS v2 (Kibana v6.2+) and publish as a 3.4MB TopoJSON file.

jsanz:
fixes #179 
fixes #164 

This PR updates the World Countries dataset with a more detailed GeoJSON and TopoJSON datasets. It also updates the Admin Regions dataset to include country data.

The size of the World Countries dataset is mostly controlled by the `interval` parameter of the `simplify` command for Mapshaper. Current intervals produce a 1.8MB GeoJSON and a 2.2MB TopoJSON (uncompressed).

The `README.md` file of the `sources/world` folder go in detail also on how to query Wikidata to generate a CSV file with population and area for each ISO2 code that is joined with Mapshaper. 

This is the visual result for the layer compared with the current production dataset, including actual download size.

image


jsanz:
An option we can also check is to replace Wikidata area/pop by [World Bank](https://databank.worldbank.org/reports.aspx?source=2&series=SP.POP.TOTL&country=) data. In the exported datasets the ISO3 code is added so this should be quite straight forward. A clear benefit is a well-defined origin of information, available for several years, also in an [Open License](https://datacatalog.worldbank.org/public-licenses#cc-by), and with many more indicators available.

@nickpeihl let me know if you think we should explore this path and I'll revert to a Draft PR.

nickpeihl:
> An option we can also check is to replace Wikidata area/pop by [World Bank](https://databank.worldbank.org/reports.aspx?source=2&series=SP.POP.TOTL&country=) data. In the exported datasets the ISO3 code is added so this should be quite straight forward. A clear benefit is a well-defined origin of information, available for several years, also in an [Open License](https://datacatalog.worldbank.org/public-licenses#cc-by), and with many more indicators available.
> 
> @nickpeihl let me know if you think we should explore this path and I'll revert to a Draft PR.

I think it's worth exploring sourcing the popluation data from the World Bank as an authoritative source.

jsanz:
Admin regions dataset needs some further work, there are some missing ISO3 codes

```
cat data/admin_regions_lvl2_v1.geo.json | jq -r '.features[] | select( .properties.country_iso3_code == "" ) | .properties.country_iso2_code ' | uniq | wc -l
```

On the other hand the World Bank offers a convenient API to get data per metric we can later join using the current Mapshaper workflow.

```
export METRIC="SP.POP.TOTL" &&\
echo "iso,measure" > ${METRIC}.csv &&\
  curl -s "http://api.worldbank.org/v2/country/all/indicator/${METRIC}?format=json&per_page=300&date=2018" |\
  jq -cr '.[1][] | select(.countryiso3code != "") | select(.value > 0) | .countryiso3code + "," + (.value | tostring)' |\
  sort >> ${METRIC}.csv
```

sources/world/README.md
```diff
@@ -1,45 +1,115 @@
...
+ ?id wdt:P297 ?iso2;
+     wdt:P1082 ?pop_value.
+}
+GROUP BY ?iso2
+ORDER BY ?iso2
+```
+
+### Mapshaper
+
+Dissolve the regions dataset by `country_iso2_code` and join with the Wikidata CSV to generate both the GeoJSON and TopoJSON outputs, each with different simplification thresholds to generate around 2MB datasets.
```
nickpeihl, line 88: Good idea to dissolve the admin regions. The Natural Earth [Admin0 Countries layer](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-0-countries/) includes some unconventional boundaries such as the Baikonur Cosmodrome.

package.json
```diff
@@ -35,6 +35,7 @@
     "eslint-plugin-mocha": "^6.3.0",
     "eslint-plugin-prefer-object-spread": "^1.2.1",
     "eslint-plugin-react": "^7.19.0",
+    "geojson-rewind": "^0.3.1",
```
nickpeihl, line 38: Nice catch.

nickpeihl (commented):
Nice work. I did a preliminary review of the data.

The updated world countries dataset is missing the following ISO-2 codes that are in the current production dataset. I suspect many of these countries have no subdivisions in the admin regions dataset. 
```
"AX"
"BQ"
"BV"
"CC"
"CX"
"GF"
"GP"
"MQ"
"RE"
"SJ"
"YT"
```

You can see this for yourself with `comm -23 <(git show master:data/world_countries_v1.geo.json | jq '.features[].properties.iso2' | sort | uniq) <(jq '.features[].properties.iso2' < data/world_countries_v1.geo.json | sort | uniq)`

As an alternative, we could use the Admin 0 Countries layer from Natural Earth. The NE Countries and NE States Provinces layers use the same boundary lines for countries. Though there might be some minor differences between the EMS layers depending on the mapshaper simplification settings we use.

sources/world/README.md
jsanz: Yeah, I think it makes sense to derive the World Countries from the Regions dataset since we have better control of the generation of this dataset.

jsanz:
@nickpeihl thanks for the review!

With last changes we have a complete World Countries dataset with all records with area and population except Antarctica and British Indian Ocean Territory. What I did is to fill the gaps from the World Bank dataset (for example both Sudan and South Sudan are empty) with the remaining information from Wikidata. I also added a `Makefile` to run the full process.

But **then** :sweat_smile:  I checked for the CIA Factbook and I found [this project](https://github.com/iancoleman/cia_world_factbook_api) that parses the website and offers a single JSON with data from countries. The ISO 3166 code is available on the Internet section.

```
cat factbook.json \| 
jq -c '.countries | to_entries | .[] | { name: .value.data.name, pop:  .value.data.people.population.total, pop_date: .value.data.people.population.date, code: (.value.data.communications.internet.country_code[1:3]) }' 
```

I like WB data because we don't rely on a scrapped HTML to source our data, but let me know if CIA Factbook is still interesting in exchange of WB or Wikidata current usage.

sources/world/README.md
```diff
@@ -1,45 +1,144 @@
-# Administrative Divisions
+# World datasets
+
+**Note** commands assume a copy of the [Natural Earth Vector repository](https://github.com/nvkelso/natural-earth-vector) is located on this folder as `naturalearth`.
```
nickpeihl, line 3: This is not quite accurate as some of the Natural Earth States and Provinces are incorrect. I wonder if we should publish our own fork of Natural Earth with our corrections?

sources/world/README.md
```diff
@@ -1,45 +1,144 @@
...
+
+After this process, there are still some records without area or population, we will complete the dataset with Wikidata information. To get the list of the missing countries we will filter using `csvsql` from the `csvkit` package over a finished dataset.
+
+* Get the country iso codes from the regions dataset
+```
+cat ../../data/admin_regions_lvl2_v1.geo.json |\
+  jq -r '.features[] | .properties.country_iso3_code + "," + .properties.country_iso2_code' |\
+  uniq | sort > /tmp/isos.csv
+```
+
```
nickpeihl, line 86: Out of scope for this PR, but maybe in a separate PR we should consider creating a Makefile to run the commands to generate these datasets?

sources/world/world_countries_v2.hjson
```diff
@@ -39,6 +55,16 @@
       name: name
       desc: Country name
     }
+    {
+      type: property
```
nickpeihl, line 59: I think we're introducing a new data type here. I suggest `metric`. Otherwise, population and area will show up as join fields in Elastic Maps.

sources/world/world_countries_v7.hjson
```diff
@@ -0,0 +1,235 @@
...
+      name: iso3
+      desc: Three letter abbreviation
+    }
+    {
+      type: property
+      name: name
+      desc: Country name
+    }
+    {
+      type: property
```
nickpeihl, line 59: Same as above
> I think we're introducing a new data type here. I suggest metric. Otherwise, population and area will show up as join fields in Elastic Maps.

nickpeihl (commented):
1. > I like WB data because we don't rely on a scrapped HTML to source our data, but let me know if CIA Factbook is still interesting in exchange of WB or Wikidata current usage.

I agree. WB has a proper authoritative API which I prefer.



2. I also see that my comments [here](https://github.com/elastic/ems-file-service/pull/188#discussion_r493034314) and [here](https://github.com/elastic/ems-file-service/pull/188#pullrequestreview-493876196) are antithetical. I'm starting to reconsider using the admin regions to create the world countries because we are deleting ISO codes that previously existed. 

3. We should make sure metric fields like `population` and `area` don't show as join fields in Elastic Maps and region maps in older releases of Kibana.

sources/world/world_countries_v2.hjson
jsanz: What's the difference then between `id` and `property` types?

sources/world/README.md
jsanz: yeah, I've been thinking about that as well, or at least to place a copy of the regions dataset somewhere in this repo to make it self contained.

sources/world/README.md
jsanz: I already added a [Makefile](https://github.com/elastic/ems-file-service/pull/188/files/aaa845ee60252c7a2b0507f7e51ead82a716905d#diff-34a562d4e8165b7c9ca64cade71f884b) to this PR :sweat_smile:

jsanz:
> I also see that my comments here and here are antithetical. I'm starting to reconsider using the admin regions to create the world countries because we are deleting ISO codes that previously existed.

From your previous https://github.com/elastic/ems-file-service/pull/188#pullrequestreview-493876196, these are the removed ISO codes from the PR data, all part of overseas territories from different countries:

ISO| Name | Notes
--|--|--|
AX | Åland Islands | Finland 
BQ | Caribbean Netherlands | Netherlands
BV | Bouvet Island | Norway
CC | Cocos (Keeling) Islands | Australia
CX | Christmas Island | Australia
GF | French Guiana | France
GP | Guadeloupe | France
MQ | Martinique | France
RE | Réunion | France
SJ | Svalbard and Jan Mayen | Norway
YT | Mayotte | France

sources/world/world_countries_v2.hjson
nickpeihl: `id` types should be used for join fields, they are usually ISO codes or other statistical codes. `property` types should be only used in tooltips, because they are usually country or region names (unsuitable for joining). However, for legacy reasons, Kibana still includes all field types in the join fields in Maps and the legacy viz.

sources/world/README.md
nickpeihl: 🙈 oops. I didn't see that. That's fantastic!

jsanz:
@nickpeihl in f4129effd1fe9eb76fd24f0de8f078bc4fe9f1ed I've moved the dataset to the `world` folder and updated instructions and so on. I've tested this branch on Kibana and apart from the well-known issue of these fields being offered for JOINs on EMS folders, everything seems to be working fine.

I still want to check for those ISO codes mentioned earlier today to see if they show up in the GeoIP database. I'll do that next week.

sources/world/Makefile
```diff
@@ -1,38 +1,45 @@
...
+  node duplicate-regions.js -v \
+    ../sources/world/duplicated-regions.csv \
+    /tmp/ne_10m_admin_1_states_provinces.geo.json \
+    /tmp/additional-regions.geo.json
+
+../../data/admin_regions_lvl2_v2.topo.json: /tmp/ne_10m_admin_1_states_provinces.topo.json /tmp/additional-regions.geo.json ## Regions GeoJSON
+	mapshaper -i   combine-files \
+  /tmp/ne_10m_admin_1_states_provinces.topo.json /tmp/additional-regions.geo.json \
   -merge-layers force \
+  -dissolve "iso_3166_2" copy-fields="iso_3166_2,name,iso_a2,adm0_a3,admin" \
```
jsanz: @nickpeihl this is the main part of my last change. I removed the ad-hoc management of Taiwan and moved all that logic to the `duplicated-regions.js` script that takes a CSV with the ISO codes that you want to extract.

Later on, the `admin_regions_lvl2_v2.topo.json` target takes the output and the original NE dataset and merges them, and as you did, there is a `-dissolve` for the regions to merge the origin regions that are a single region in the duplicated zone, as for Taiwan or Caribbean Netherlands.

sources/world/duplicated-regions.csv
```diff
@@ -0,0 +1,38 @@
...
+NO-X01~,Bouvet Island,BV,Bouvet Island,BV,BVT
+-99-X15~,Cocos (Keeling) Islands,-99-X15~,Cocos (Keeling) Islands,CC,CCK
+-99-X14~,Christmas Island,-99-X14~,Christmas Island,CX,CXR
+FR-GF,French Guiana,GF-GF,French Guiana,GF,GUF
+FR-GP,Guadeloupe,GP-GP,Guadeloupe,GP,GLP
+FR-MQ,Martinique,MQ-MQ,Martinique,MQ,MQT
+FR-RE,Réunion,RE-RE,Réunion,RE,REU
+FR-YT,Mayotte,YT-YT,Mayotte,YT,MYT
+NO-21,Svalbard,SJ-21,Svalbard and Jan Mayen,SJ,SJM
+NO-22,Jan Mayen,SJ-22,Svalbard and Jan Mayen,SJ,SJM
```
jsanz, line 38: Like here for these two regions, there are a few more _made-up_ output region codes. They were needed for the `dissolve` process for the regions dataset.

sources/world/world_countries_v7.hjson
```diff
@@ -0,0 +1,235 @@
...
+      name: population
+      desc: Population
+    }
+    {
+      type: property
```
nickpeihl, line 64:
```suggestion
      type: metric
```

sources/world/world_countries_v2.hjson
```diff
@@ -39,6 +55,16 @@
...
+      name: population
+      desc: Population
+    }
+    {
+      type: property
+      name: area
+      desc: Area
+    }
```
nickpeihl, line 67:
```suggestion

```
We'll never support metric fields in older versions of Kibana that use this version of World Countries. So I think we can remove these fields from the schema.

nickpeihl (changes requested):
I added a few thoughts from a code review.

Sorry, I still need more time to review the datasets.

sources/world/Makefile
```diff
@@ -0,0 +1,96 @@
...
+  -rename-layers data \
+  -sort this.properties.region_iso_code \
+  -o format=topojson prettify ../../data/admin_regions_lvl2_v2.topo.json
+
+../../data/admin_regions_lvl2_v1.geo.json: ../../data/admin_regions_lvl2_v2.topo.json ## Regions TopoJSON
+	mapshaper -i ../../data/admin_regions_lvl2_v2.topo.json \
+  -simplify visvalingam interval=1000 keep-shapes \
+  -sort this.properties.region_iso_code \
+  -o rfc7946 format=geojson prettify ../../data/admin_regions_lvl2_v1.geo.json
+	node ../../scripts/clean-geom.js -v ../../data/admin_regions_lvl2_v1.geo.json
```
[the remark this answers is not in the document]
jsanz: The `-simplify` generates a few invalid geometries.

I will add an additional `mapshaper` step to generate a prettified, they are useful for diffs as #190 proved :smile:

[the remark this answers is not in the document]
jsanz: BTW, I tried to modify `JSON.stringify` on `clean-up.js`, but mapshaper's pretty-printing is adapted for GeoJSON and much better for reading.

sources/world/world_countries_v2.hjson
jsanz: I'll remove the fields for the `geo.json` output, thanks

sources/world/Makefile
```diff
@@ -0,0 +1,96 @@
...
+  -o format=geojson /tmp/ne_10m_admin_1_states_provinces.geo.json
+
+/tmp/additional-regions.geo.json: /tmp/ne_10m_admin_1_states_provinces.geo.json ./duplicated-regions.csv
+	cd ../../scripts && \
+  node duplicate-regions.js -v \
+    ../sources/world/duplicated-regions.csv \
+    /tmp/ne_10m_admin_1_states_provinces.geo.json \
+    /tmp/additional-regions.geo.json
+
+../../data/admin_regions_lvl2_v2.topo.json: /tmp/ne_10m_admin_1_states_provinces.topo.json /tmp/additional-regions.geo.json ## Regions GeoJSON
```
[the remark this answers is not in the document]
jsanz: I think it's better to use explicit filenames on targets, I'll add `.PHONY: clean` for completeness.

jsanz:
With the last changes, the metric fields are omitted by Kibana as expected.

image

sources/world/Makefile
```diff
@@ -0,0 +1,103 @@
...
+
+/tmp/missing-population.csv: /tmp/isos.csv /tmp/population.csv ## Missing population from the World Bank dataset
+	csvsql -H  --query 'select isos.b from isos left join population on isos.a = population.a where population.b is null ' /tmp/isos.csv /tmp/population.csv |\
+  tail +2 > /tmp/missing-population.csv
+
+/tmp/wikidata.csv: /tmp/missing-area.csv /tmp/missing-population.csv ## Wikidata area and population for the missing records
+	codes=$$(sort /tmp/missing-area.csv /tmp/missing-population.csv | uniq | awk '{ print "\""$$0"\""}') && curl -H "Accept: text/csv" -sG 'https://query.wikidata.org/sparql' --data-urlencode query="SELECT ?iso2 (MAX(?p) AS ?population) (MAX(?a) AS ?area) WHERE { VALUES ?iso2 { $${codes}  } ?id wdt:P297 ?iso2; wdt:P2046 ?a; wdt:P1082 ?p. } GROUP BY ?iso2 ORDER BY ?iso2" > /tmp/wikidata.csv
+
+$(DATA_DIR)/world_countries_v7.topo.json: $(DATA_DIR)/admin_regions_lvl2_v1.geo.json /tmp/area.csv /tmp/population.csv /tmp/wikidata.csv ## World Countries TopoJSON
+	mapshaper -i $(DATA_DIR)/admin_regions_lvl2_v1.geo.json \
```
nickpeihl, line 82: I think we may have to dissolve from an intermediary admin_regions dataset that does not include the additional regions. Because now we are getting overlapping polygons some countries like French Guiana (as part of FR) and Taiwan (as part of CN). This could be misrepresenting data in our choropleth maps. 

To clarify, I'm OK with adding the additional regions in the admin_regions dataset, but not in the world countries dataset.

nickpeihl (changes requested):

sources/world/Makefile
nickpeihl: Maybe I'm being overly cautious. I just tested this layer with a sample dataset. It is possible to page through tooltip results in Elastic Maps. Here's a GIF of what I saw. French Guiana shows up as red because the FR polygon appears first. So the blue polygon for GF does not appear. What do you think?

Oct-01-2020 13-02-58

This might not be the most common case. I imagine China overlapping Taiwan would be more common.

jsanz: This is mostly problematic for Taiwan, and until we have a better way to deal with disputed borders, I think we can stick to our current representation of the world countries.

In a483 I've added a filter to filter the appropriate polygons from the regions dataset before dissolving to the countries. It takes the codes making two SQL queries to the input CSV.

jsanz:
@nickpeihl sorry for the extra commits, I did a rebase from master to include the Italy Provinces change but I should have merged :sweat: 

I've also updated `feature-layers` branch to align with `master` and this PR is now against `feature-layers` so we can deploy this to staging environment for further testing after merge (or maybe better squash in this case).

nickpeihl (approved):
final changes lgtm! thanks.

sources/world/Makefile
```diff
@@ -0,0 +1,96 @@
+.DEFAULT_GOAL := ../../data/world_countries_v1.geo.json
```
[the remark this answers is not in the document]
jsanz: forgot to mention here that unfortunately variables are not expanded for target names or dependencies on Makefiles, so I used it only inside commands.

PR #190, opened 2020-09-29
Add IT-SD code to Italy Provinces dataset

### LINKED ISSUE #189: Missing province code
The ISO 3166-2 code for the "Province of South Sardinia" province in the Italy Provinces map is missing:

{ properties":{"istat":"111","iso_3166_2":"NULL","label_en":"Province of South Sardinia","label_it":"provincia del Sud Sardegna"}

it should be instead:

{ properties":{"istat":"111","iso_3166_2":"IT-SD","label_en":"Province of South Sardinia","label_it":"provincia del Sud Sardegna"}

jsanz:
Fixes #189 

South Sardinia ISO code was added to the revision of Italy provinces ([provinces_v2.htjson](https://github.com/elastic/ems-file-service/blob/master/sources/it/provinces_v2.hjson)) that will be available in the next major, but never _backported_ to the older version.

This is just a replacement of the ISO code, without running the full Sophox procedure.

### DIFF
--- a/data/italy_provinces_v1.geo.json
+++ b/data/italy_provinces_v1.geo.json
@@ -1449,7 +1449,7 @@
   },
   "properties": {
     "istat": "111",
-    "iso_3166_2": "NULL",
+    "iso_3166_2": "IT-SD",
     "label_en": "Province of South Sardinia",
     "label_it": "provincia del Sud Sardegna"
   },


nickpeihl (approved):
lgtm! thanks.

PR #195, opened 2020-10-20
Fixed geometries for USA, Cyprus, and Netherlands

jsanz:
Fixes #194 

This PR updates the regions and countries datasets to fix the missing geometries for US, Cyprus, and the  Netherlands.

PR #192, opened 2020-10-07
GeoJSON datasets

jsanz:
Fixes #191 

This PR includes the following new GeoJSON data sources, with the same precision that their already existing TopoJSON versions.

* `admin_regions_lvl2_v2.geo.json` 25MB
* ` usa_counties_v2.geo.json` 2MB
* `usa_zip_codes_v7.geo.json` 11MB
* `world_countries_v7.geo.json` 11MB

Changes in existing admin regions and world countries are related to a small fix in the data pipeline to include proper IDs that I did as part of improving the `Makefile`. 

For the World Countries, the new datasets are only available from `7.10`, offering to the older versions the current published low precision dataset.

PR #196, opened 2020-10-21
World Countries and Administrative Regions update

jsanz:
Opening the final PR to take into production the changes in the datasets developed on PRs #188 #192 and #195 that improve our World Countries and Admin Regions dataset.

Some remarks:

* The Administrative Regions dataset now includes country name and ISO code
* The new World Countries dataset:
  * is much more detailed but only available from 7.10.
  * contains population and area metric fields for future consumption by Kibana to help in data normalization processes.
  * is derived from the Administrative Regions dataset, merging regions from the same country
* The manifest for 7.10 includes both TopoJSON and GeoJSON versions of both datasets. TopoJSON is aimed for visualization purposes (Kibana) while GeoJSON is aimed for more generic consumption or analysis, like ingestion into Elasticsearch.

Check below a quick clip comparing the datasets between `7.8` and `7.10-SNAPSHOT`.

These changes can be tested from the staging version in Kibana configuration or from [EMS landing page](https://maps.elastic.co/v7.10/?manifest=testing#file/world_countries).

```yaml
map.emsFileApiUrl: https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev
```

### 7.10

 

cc. @thomasneirynck @nickpeihl


nickpeihl (approved):
lgtm! tested in Kibana 7.10 and 7.9 with and without proxying in Kibana.

We probably discussed it, but I can't remember. What was our reasoning for exposing the country codes in the administrative regions data? Tooltips? It doesn't look like we can filter on country codes on EMS Boundaries in Maps and I don't know why anyone would want to use them in joins.

jsanz:
> We probably discussed it, but I can't remember. What was our reasoning for exposing in the manifest the country codes in the administrative regions data? Tooltips? It doesn't look like we can filter on country codes on EMS Boundaries in Maps and I don't know why anyone would want to use them in joins.

I don't think we discussed it but my reasoning is that could be useful for anyone downloading it and ingesting it in the cluster. I think they can also be a good addition to any map tooltip showing regions

image


Finally, I also liked the idea of an explicit field linking both datasets instead of relying on the region iso2 prefix, but this is more speculative.

Still, if that's not something we'd like to offer we could simply remove the columns in the data processing after generating the countries dataset.

thomasneirynck (approved):
thx for all the updates. these are becoming really valuable, high quality datasets (and not just for Kibana users).

as discussed with @jsanz, I'd consider removing the metric-fields from the actual data (topo and geojson files) in the world-countries. Until there is support for these in Kibana ( in the UX (https://github.com/elastic/kibana/issues/28367) and with supporting math-scripting (https://github.com/elastic/kibana/issues/28265)) this is "phantom"-data that an end-user will not know is there.

jsanz:
At ad846ee I've left the previous `make` targets and duplicated them without the metric processing so now the default target does not query World Bank or wikidata APIs to attach any area or population information. This will be easily reverted when we have metric support at Kibana and we are ready to advertise these type of fields.

nickpeihl:
> Still, if that's not something we'd like to offer we could simply remove the columns in the data processing after generating the countries dataset.
I think we should leave the columns in the data. I was more concerned about [listing those fields in the manifest](https://github.com/elastic/ems-file-service/pull/196/files#diff-3a38985184f4b534c34a3d6ae97604234702e03b63a3ceefdd02dd9d3eff3baeR45-R59) since they aren't very useful as join fields. But they are probably useful for tooltips so I think this is fine.

PR #197, opened 2020-10-27
Fix Lithuania and China data issues

jsanz:
This PR fixes missing Lithuania and CHina data from the administrative regions and subsequently the world countries dataset.

Fixes can be observed in [staging](https://maps.elastic.co/v7.10/?manifest=testing#file/administrative_regions_lvl2).

I still see a few missing region names, all corresponding with small islands, but we are probably fine on that front since they may not have proper iso codes according to their current values

image

**Note**: I'm not totally sure why previous commits are being picked again on this PR, probably because I didn't merge from master before pushing 12f4d4b and I did it later at 5dad78c (my bad)


nickpeihl (approved):
lgtm! thanks.

will merge and publish this after ci completes.

PR #199, opened 2020-10-28
Fix error on countries source id

### LINKED ISSUE #198: World Countries not working in Version 2
Today we found an issue in our Region Maps Visualizations, they aren't able to load and promps the following error,

"Error downloading vector data
Cannot download https://vector.maps.elastic.co/blob/5659313586569216?elastic_tile_service_tos=agree&my_app_version=YYY&license=XXX file. Please ensure the CORS configuration of the server permits requests from the Kibana application on this host."

After some checks, I thought it could be a API problem in maps.elastic.co
    For one hand, maps.elastic.co v1 works properly for "World Countries" layer:
        https://maps.elastic.co/index.html#file/world_countries
    But it seems to be unavailable for v2 version:
        https://maps.elastic.co/v2/index.html#file/World%20Countries
 
   
If we configure the Region Maps Visualization to another layer, Belgium Regions, in v2 seems to work properly:
    https://maps.elastic.co/v2/index.html#file/Belgium%20Regions

jsanz:
Fixes #198 

When creating the new World Countries layer I used the same source identifier, which made the v2 manifest to point to a blob that contains the TopoJSON version of the dataset.

To check this PR works the simplest way is to see that the staging manifest for `v2` points to the GeoJSON dataset

```
$ curl -s --compressed https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev/v2/manifest\?elastic_tile_service_tos\=agree\&my_app_name\=ems-landing\&my_app_version\=7.2.0\&license\=643c1faf-80fc-4ab0-9323-4d9bd11f4bbc | jq '.layers[0].url'
"https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev/blob/5659313586569216?elastic_tile_service_tos=agree"
```

```
$ curl -s --compressed "https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev/blob/5659313586569216?elastic_tile_service_tos=agree" | jq '.features[0].properties'
{
  "name": "Antigua and Barbuda",
  "iso2": "AG",
  "iso3": "ATG"
}
```


sources/world/world_countries_v2.hjson
```diff
@@ -202,5 +202,5 @@
   ttl: 108000
   weight: 2
   createdAt: "2017-04-26T17:12:15.978370"
-  id: 5659313586569216
+  id: 5659313586569220
```
nickpeihl, line 205:
```suggestion
  id: 5659313586569216
```

I'm not sure it matters, but I'd suggest not changing the old `id`. It might be saved in older visualizations.

nickpeihl (approved):
just one nit, but otherwise lgtm!

Issue #206, opened 2021-03-29
[Meta] EMS 7.13


nickpeihl:
It's time to prep EMS v7.13. We'll add corresponding issues to each item below.
 
- [x] File Service https://github.com/elastic/ems-file-service/pull/207
- [x] Tile Service
- [x] ems-client https://github.com/elastic/ems-client/pull/59
- [x] ems-landing-page  https://github.com/elastic/ems-landing-page/pull/329
- [x] bump ems-client in kibana 
- [x] update ems-landing-page link in kibana

jsanz:
EMS File service 7.13 deployed to production https://kibana-ci.elastic.co/job/elastic+ems-file-service+deploy/15/

PR #207, opened 2021-03-29
Release EMS 7.13

nickpeihl:
Part of #206.

This PR will set up the Elastic Maps Service release for 7.13. 

Note: This also adds releases up to 7.16 with the caveat that future releases of Elastic Maps Service are in development and may change at any time.

### DIFF
--- a/scripts/constants.js
+++ b/scripts/constants.js
@@ -5,9 +5,26 @@
  */
 
 module.exports = Object.freeze({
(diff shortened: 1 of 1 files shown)


jsanz (approved):
Changes in code for future releases shouldn't be an issue as they won't be accessed yet, so :+1: on my end. Thanks!!

PR #212, opened 2021-05-05
Added a new field for world countries dataset with the ISO numeric code

### LINKED ISSUE #211: Support for ISO 3166-1 numeric
Could you please the ISO 3166-1 numeric country codes to world countries data.

jsanz:
Fixes #211 

This PR adds a new table in the `sources/world` folder with the list of iso 2, 3, and numeric country codes, adding the numeric code to the world countries GeoJSON and TopoJSON targets in the Makefile.

CSV is taken from <https://www.iso.org/obp> website, more details in ISO [country codes landing page](https://www.iso.org/iso-3166-country-codes.html).


sources/world/Makefile
```diff
@@ -123,24 +123,25 @@ clean: ## Remove output files
 
 
 # Data without metrics
-../../data/world_countries_v7.geo.json: ../../data/admin_regions_lvl2_v2.geo.json /tmp/additional-regions.geo.json ## World Countries GeoJSON full size
+../../data/world_countries_v7.geo.json: ../../data/admin_regions_lvl2_v2.geo.json /tmp/additional-regions.geo.json ./country-codes.csv ## World Countries GeoJSON full size
 	codesA=$$(csvsql --query 'select distinct in_region_iso from "duplicated-regions" where not keep' duplicated-regions.csv | tail +2 | sed -e 's/^/"/g' -e 's/$$/"/g' | paste -sd "," -) && codesB=$$(csvsql --query 'select distinct out_region_iso2 as iso from "duplicated-regions" where keep' duplicated-regions.csv | tail +2 | sed -e 's/^/"/g' -e 's/$$/"/g' | paste -sd "," -) && mapshaper -i $(DATA_DIR)/admin_regions_lvl2_v2.geo.json \
   -filter "[$${codesA},$${codesB}].indexOf(region_iso_code) == -1" \
   -dissolve 'country_iso2_code' 'copy-fields=country_iso3_code,country_name' \
   -rename-fields "iso2=country_iso2_code,iso3=country_iso3_code,name=country_name" \
+  -join ./country-codes.csv "keys=iso2,iso2" "fields=iso_numeric" \
```
nickpeihl, line 131: I think we need to add the `iso_numeric` codes as strings rather than integers. Maybe using the `string-fields` option in the [Mapshaper `-join`](https://github.com/mbloch/mapshaper/wiki/Command-Reference#-join)?

sources/world/country-codes.csv
```diff
@@ -0,0 +1,250 @@
+iso2,iso3,iso_numeric
+AF,AFG,4
```
nickpeihl, line 2: The [ISO numeric codes are three digits](https://en.wikipedia.org/wiki/ISO_3166-1_numeric) so some countries will need leading zeros. 

For example: 
```suggestion
AF,AFG,004
```

data/world_countries_v7.topo.json
```diff
@@ -15,6 +15,7 @@
           "properties": {
             "iso2": "AD",
             "iso3": "AND",
+            "iso_numeric": 20,
```
nickpeihl, line 18: I think these should be strings.

For example
```suggestion
            "iso_numeric": "020",
```

nickpeihl (changes requested):

sources/world/country-codes.csv
jsanz: Yep, sorry about that, thanks for the pointer!

jsanz:
@nickpeihl I've addressed the leading zeros issue.

I also removed unnecessary attribution to World Bank and Wikidata that was introduced when metrics were added (I guess?) and added attribution to OSM as we have for the Admin Regions dataset since world borders are derived from Admin Regions.

nickpeihl (approved):
lgtm!

Issue #210, opened 2021-03-30
Add well-known column aliases to fields in manifest


nickpeihl:
Similar to #209, clients may want to autosuggest term joins to a field based on well-known column aliases. ECS fields are one example, but we could consider others. I think this would be an array of regex patterns that could be matched to a field in an Elasticsearch query result. See my example vector manifest below. See https://github.com/elastic/kibana/pull/94969/files.

cc @thomasneirynck @jsanz

```json
{
  "version": "7.13",
  "layers": [
    {
      "layer_id": "world_countries",
      "created_at": "2020-10-28T16:16:08.720286",
      "attribution": [ ... ],
      "formats": [ ... ],
      "fields": [
        {
          "type": "id",
          "id": "iso2",
          "label": { ... },
          "alias": [
            /(geo\.){0,}country_iso_code$/,  <!-- ECS fields -->
            /iso(-|_)*(a)*(2)*/i
          ]
        },
        {
          "type": "id",
          "id": "iso3",
          "label": { ... },
          "alias": [
            /iso(-|_)*(a)*(3)*/i
          ]
        },
        {
          "type": "property",
          "id": "name",
          "label": { ... },
          "alias": [
            /(geo\.){0,}country_name$/, <!-- ECS fields -->
            /^country/i
          ]
        }, ... ]
...
```

nickpeihl:
Closed by #215

PR #215, opened 2021-05-10
Add alias, regex, and values properties to some fields

### LINKED ISSUE #209: Add regex pattern for fields in manifest
We might want to consider adding a regex pattern property to the vector file fields. This could be useful for clients that want to automatically suggest a field matching a pattern without downloading the full file. See https://github.com/elastic/kibana/pull/94969. 

cc @jsanz @thomasneirynck 

An example is below.

```json
{
  "version": "7.13",
  "layers": [
    {
      "layer_id": "world_countries",
      "created_at": "2020-10-28T16:16:08.720286",
      "attribution": [ ... ],
      "formats": [ ... ],
      "fields": [
        {
          "type": "id",
          "id": "iso2",
          "regexp": /^[A-Z]{2}$/,
          "label": { ... }
        },
        {
          "type": "id",
          "id": "iso3",
          "regexp": /^[A-Z]{3}$/,
          "label": { ... }
        },
        {
          "type": "property",
          "id": "name",
          "regexp": /^.*$/,
          "label": { ... }
        }, ... ]
...
```

nickpeihl:
Fixes #209 and #210

This PR adds support for optional `regex`, `alias`, and values properties on fields in the manifest. The `regex` property is a regular expression pattern that matches all field values in the vector file.

The `alias` property is an array of regular expressions that can be considered aliases for the field name. This should make it easier to suggest a matching field for joining data to the vectors, such as [ECS geo fields](https://www.elastic.co/guide/en/ecs/current/ecs-geo.html).

The `fields` property is an array of field values for fields of type `id` parsed from the GeoJSON files. This should make it easier for clients to autosuggest a layer based on the manifest without downloading the entire GeoJSON file.

I bumped the AJV dependency because it adds JSON schema support for string formats, particularly strings that match regular expressions.

Since this is new functionality, I enabled these properties only on EMS v7.14 and above.


jsanz (approved):
Tested and built locally, I can't see anything wrong, nice job @nickpeihl :heart_eyes: 

I have two questions:

* Should we check the actual data content based on the `regex` field (if exists) to ensure it is valid?
* Should we add the regex to all sources with custom identifiers? This could go on a separate PR. I did a small check searching for additional identifiers :point_down: 

<details>

```text
➜ for i in $(find . -name "*.hjson"); do result=$(cat $i | hjson -j | jq -c '.fieldMapping[] | select(.type == "id") | select (.name != "iso_3166_2") | select(.regex == null) '); if [ -n "$result" ]; then echo "$i\r\n$result\r\n"; fi ; done
./cl/chile-regions.hjson
{"type":"id","name":"cut","desc":"CUT code"}

./us/states_v1.hjson
{"type":"id","name":"postal","desc":"Two letter abbreviation"}

./us/states_v8.hjson
{"type":"id","name":"postal","desc":"Two letter abbreviation"}

./us/counties.hjson
{"type":"id","name":"fips","desc":"FIPS code"}
{"type":"id","name":"gnis","desc":"GNIS ID"}

./us/zip_codes_v2.hjson
{"type":"id","name":"zip","desc":"5-digit zip code"}

./uk/subdivisions.hjson
{"type":"id","name":"gss","desc":"GSS code"}

./ie/ireland-counties.hjson
{"type":"id","name":"logainm","desc":"Logainm identifier"}

./cn/provinces.hjson
{"type":"id","name":"division_code","desc":"China administrative division code"}

./world/world_countries_v2.hjson
{"type":"id","name":"iso2","desc":"Two letter abbreviation"}
{"type":"id","name":"iso3","desc":"Three letter abbreviation"}

./world/admin_regions_lvl2_v1.hjson
{"type":"id","name":"region_iso_code","desc":"Region ISO code"}

./fr/departments_v1.hjson
{"type":"id","name":"insee","desc":"INSEE department identifier"}

./fr/departments_v7.hjson
{"type":"id","name":"insee","desc":"INSEE department identifier"}

./jp/prefectures.hjson
{"type":"id","name":"dantai","desc":"Dantai administrative division code"}

./ch/switzerland-cantons.hjson
{"type":"id","name":"canton_code","desc":"Canton Code"}

./kr/south-korea-municipalities.hjson
{"type":"id","name":"sigungu_cd","desc":"Sigungu Code (시군구코드)"}

./it/regions.hjson
{"type":"id","name":"istat_id","desc":"ISTAT ID"}

./it/provinces_v1.hjson
{"type":"id","name":"istat","desc":"ISTAT ID"}

./it/provinces_v2.hjson
{"type":"id","name":"istat","desc":"ISTAT ID"}
```
</details>

sources/world/world_countries_v7.hjson
```diff
@@ -44,16 +44,24 @@
       type: id
       name: iso2
       desc: Two letter abbreviation
+      regex: ^[A-Z]{2}$
+      alias: [
+        (geo\.){0,}country_iso_code$
```
thomasneirynck, line 49: can we add a 2nd entry? (and add it to the `name` field

`(country|countries)`

This was raised here https://github.com/elastic/kibana/pull/99553.

It would also help to validate the functionality with multiple aliases. 

e.g. A field with very tight pattern-validation (e.g. `iso2`),  can be more "lenient" with field-aliases, since the we can explicitly validate the values.

It's something that needs to be handled on the Kibana side (still iterating).

thomasneirynck (commented):
this is great.

can we try and get in more examples of multiple aliases?

sources/world/world_countries_v7.hjson
nickpeihl: > can we add a 2nd entry? (and add it to the `name` field
> 
> `(country|countries)`
> 
> This was raised here [elastic/kibana#99553](https://github.com/elastic/kibana/pull/99553).
> 
> It would also help to validate the functionality with multiple aliases.
> 
> e.g. A field with very tight pattern-validation (e.g. `iso2`), can be more "lenient" with field-aliases, since the we can explicitly validate the values.
> 
> It's something that needs to be handled on the Kibana side (still iterating).

Discussed offline with Thomas and I will add the suggested aliases.

jsanz: > I added this test here to check the data. Is that what you mean?

:facepalm: yeah, that's exactly what I meant :sweat_smile:

jsanz (approved):
Thanks for adding all the new regexes! :raised_hands:

nickpeihl:
@jsanz @thomasneirynck I updated this PR to also add the field values to the manifest. Fixes issues #215 and #217.

schema/source_schema.json
```diff
@@ -185,6 +185,19 @@
...
+          "format": "regex"
+        },
+        "alias": {
+          "type": "array",
+          "items": {
+            "description": "Regular expressions that may match the field.",
+            "type": "string",
+            "format": "regex"
+          }
         }
```
jsanz, line 201: Should we have the new `values` property for the id `field`?
```suggestion
          }
        },
        "values": {
          "type": "array",
          "items": {
            "description": "Actual identifiers for the boundaries of the dataset",
            "type": "string"
          }
        }
```

jsanz (approved):
Looks good to me, just added a nit.

schema/source_schema.json
nickpeihl: Interesting idea. I hadn't considered adding `values` to the source files. I just assumed we'd only populate the `values` in the destination manifest from the geojson files.

Do you see a use case for adding a `values` array to the source files? I suppose it could be used to strictly validate the GeoJSON files, but that may be out of scope for this PR.

jsanz: Apologies Nick, I totally confused the schema of the source and the manifest :facepalm:

scripts/generate-manifest.js
```diff
@@ -186,7 +186,23 @@ function manifestLayerV6(data, hostname, opts) {
...
-  return emsFormats.filter(format => format.default)[0];
+  return emsFormats.find(format => format.default);
+}
+
+function getIdsFromFile(dataDir, file, fields) {
+  const json = JSON.parse(readFileSync(`${dataDir}/${file}`, 'utf8'));
+  const features = json.features || json.objects.data.geometries;
+  const fieldMap = {};
+  for (const {name} of fields) {
+    fieldMap[name] = new Set(); // Probably unnecessary but ensures unique ids
```
thomasneirynck, line 197: OK to have this here. 

tangent: would it make sense to add a unit tests so that every field of `"type": "id"` is a primary-key?

scripts/generate-manifest.js
```diff
@@ -186,7 +186,23 @@ function manifestLayerV6(data, hostname, opts) {
...
+  };
+  for (const feature of features) {
+    const { properties } = feature;
+    for (const {name} of fields) {
+      fieldMap[name].add(properties[name])
```
thomasneirynck, line 202: nit: but these destructerings add noise imho

   `fieldMap[name].add(feature.properties[name])`

thomasneirynck (approved):
this is killer. A significant pre-req to introduce EMS into the rest of the stack/solutions.

scripts/generate-manifest.js
nickpeihl: Good idea. I added a unit test for this, but immediately ran into issues with legacy layers. So I demoted the unit test to only run when the [`EMS_STRICT_TEST` variable is set](https://github.com/elastic/ems-file-service#usage). We can fix the legacy layers in a separate PR. 

Once we fix those layers we can promote the unit test and probably remove this `Set`.

PR #218, opened 2021-05-28
Fix iso3 and name for Tokelau

nickpeihl:
I discovered this error when I ran `EMS_STRICT_TEST=YES yarn test` which was recently updated to check the cardinality of fields of type `id`.


jsanz (approved):
:+1: 

Side note: it's great the strict settings now can find this kind of issues as well. I see that currently, we throw errors for France, Peru, Montenegro, and Slovenia. The first fails because of the lack of IDs for many departments and for the rest the lack of some labels in country languages.

Should we address these errors to enable strict checking by default in our CI?

nickpeihl:
> Side note: it's great the strict settings now can find this kind of issues as well. I see that currently, we throw errors for France, Peru, Montenegro, and Slovenia. The first fails because of the lack of IDs for many departments and for the rest the lack of some labels in country languages.
> 
> Should we address these errors to enable strict checking by default in our CI?

Yes, I think we should fix these errors in separate PRs. I don't think we would be breaking any existing visualizations by doing so.

PR #221, opened 2021-06-15
Fix Tokelau from admin regions data source

jsanz:
This PR is a fix over #218 that made the change directly in the output of the Natural Earth data pipeline. On this PR the source dataset is fixed and the output files for admin regions and world countries have the correct data for Tokelau.


nickpeihl (approved):
lgtm! thanks.

PR #229, opened 2021-10-19
Stage possible 8.x versions

nickpeihl:
Stages possible 8.x EMS releases.

**Note:** Future releases are subject to change without notice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "Data sources, vector data, and manifest generation for Elastic Maps Service",
   "main": "index.js",
   "scripts": {
-    "test": "tap test/*.js --cov",
+    "test": "tap test/*.js --cov --reporter base",
     "build": "node index.js"
   },
   "keywords": [],


scripts/constants.js
```diff
@@ -8,7 +8,7 @@ module.exports = Object.freeze({
   VECTOR_PRODUCTION_HOST: "vector.maps.elastic.co",
   VECTOR_STAGING_HOST: "vector-staging.maps.elastic.co",
   TILE_PRODUCTION_HOST: "tiles.maps.elastic.co",
-  TILE_STAGING_HOST: "tiles.maps.elstc.co",
+  TILE_STAGING_HOST: "tiles.maps.elastic.co",
```
jsanz, line 11: Was this a typo or an old URL that we don't need to use anymore? Just curious.

nickpeihl: Sorry, I forgot to mention this change in the description. This is an old URL that we don't use anymore.

jsanz:
> Thanks. I've added a reporter to the tap command. I think I prefer the base reporter, but I'm happy to iterate on this in another PR if you feel differently.

The problem with `base` reporter is that it breaks the project build with node `16` but I agree that we can iterate later. Maybe we can also specify the preferred node engine with a `.nvmrc` file.

PR #234, opened 2021-12-16
Pin the build to Node Fermium releases

jsanz:
Change the `lts` Docker image by the `fermium` (14.x) image series. We need to update our testing code to keep up with NodeJS.


thomasneirynck (approved):
code review

PR #236, opened 2021-12-20
Upgrade to tap 15 and Node 16

jsanz:
* Upgrading `tap` to `15.1.5`
  * Following deprecation notices, witching `deepEquals` and `false` methods to the new `same` and `notOk`
* Making explicit the support for node versions between `12` and `16` in package.json `engines` section
* Adding a `.nvmrc` control file to allow automatic switch to node `16` (Gallium)
* Updating the build script to use the Docker `node:gallium` image

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/gallium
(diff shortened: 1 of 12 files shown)


thomasneirynck (approved):
code review, lgtm

Issue #231, opened 2021-10-26
France Departments winding order incompatible with Vega


nickpeihl:
It looks like the coordinate winding order for France Departments is not compatible with Vega. This causes Vega to fill the entire area outside the polygon instead of the internal polygon.

Other layers in EMS such as China Provinces render correctly because the coordinate winding order is reversed. To see this change the `data[0].url.name` to `China Provinces` in the example code below.

I'm not sure if there is a way we can fix this in Vega or if we need to fix this in EMS. I'd rather see if we could fix this in Vega and just maintain the RFC7946 winding order for all EMS layers. But I also think Elastic Maps is more lenient on winding order. And GeoJSON Upload also appears to wind correctly the coordinate order.

If we fix this in EMS, we also need to check which other layers may cause winding issues with Vega.

Example Vega code
```
{
  $schema: https://vega.github.io/schema/vega/v3.0.json
  config: {
    kibana: {
      type: map
      latitude: 48
      longitude: -5
      zoom: 3
    }
  }
  data: [
    {
      name: countries
      url: {%type%: "emsfile", name: "Spain Provinces"}
      format: {type: "json", property: "features"}
    }
  ]
  marks: [
    {
      type: shape
      from: { data: "countries" }
      encode: {
        "enter": {
          "strokeWidth": {"value": 0.5},
          "stroke": {"value": "#bbb"},
          "fill": {"value": "#e5e8d3"}
        }
      }
      transform: [
        {type: "geoshape", projection: "projection"}
      ]
    }
  ]
}

```

jsanz:
We should make this a test to find any other dataset that is breaking in Vega.

jsanz:
I've been investigating this a bit more.

With [geojsonhint](https://github.com/mapbox/geojsonhint) I've checked and 61 of our current datasets have one or more warnings about the right-hand rule:

```
  1:1  warning  Polygons and MultiPolygons should follow the right-hand rule
```

[...]

PR #239, opened 2021-12-28
Adds a constant to check GeoJSON RFC7946 testing and building

jsanz:
Related to #231

This PR adds a new constant `GEOJSON_RFC7946`, set to `undefined` at this moment to not change the current results of tests and build, leaving the definition of this behavior for a follow-up PR when discussion in #231 is concluded.

When set to `true` to enforce the [RFC7946 standard](https://datatracker.ietf.org/doc/html/rfc7946#appendix-B.1) for polygon orientation in GeoJSON files, the build process will ensure that all published polygons in all data layers are counter-clockwise oriented. When testing with `EMS_STRICT_TEST` env var defined, it will report which data layers are not following the standard.

When set to `false` it will build all polygons clockwise oriented and with `EMS_STRICT_TEST` it will report the data layers that follow the RFC7946 standard and are being rewind'ed in the build process.

<details><summary>Screenshots of test executions with EMS_STRICT_TEST</summary>

### `GEOJSON_RFC7946 = undefined` 

no changes in build and test results:

undefined

---

### `GEOJSON_RFC7946 = true` 

There are 62 files that are not following the standard.

enforce

---

### `GEOJSON_RFC7946 = false`

There are 20 files that follow the standard and are consequently reported by the tests.

neglect

</details>

### DIFF
--- a/test/source.js
+++ b/test/source.js
@@ -13,6 +13,7 @@ const glob = require('glob');
 const fs = require('fs');
 const jsts = require('jsts');
 const _ = require('lodash');
+const constants = require('../scripts/constants');
 
 const ajv = new Ajv();
 addFormats(ajv);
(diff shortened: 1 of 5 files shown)


test/source.js
```diff
@@ -97,27 +98,27 @@ function validateObjectsMember(topojson, format, fieldMap, t) {
...
 }
 
-function validateGeoJSON(geojson, fieldMap, t) {
+function validateGeoJSON(geojson, fieldMap, t, fileName) {
   const reader = new jsts.io.GeoJSONReader();
   const fc = reader.read(geojson);
   const fieldsNames = fieldMap.map(f => f.name).sort();
   t.ok(fc.features.every(feat => feat.geometry.isSimple()
-  ), 'All geometries must be simple');
+  ), `All geometries from ${fileName} must be simple`);
```
nickpeihl, line 106: Adding the `fileName` is very helpful. Thank you!

nickpeihl (approved):
lgtm! this is a good implementation that will make it easy to enforce depending on the conclusion of #231

PR #243, opened 2022-03-24
Avoid adding common values on identifier fields

### LINKED ISSUE #242: Remove region identifiers with values that are too common
Related with elastic/kibana#125985

Our vector manifests ship with region identifiers to help consumers to match potential datasets that could be joined with our boundaries. 

We are seeing a root for false positives that could be fixed at the EMS File Service side related with identifier values that are too common.

For example, France departments have the `ISO` field that is very distinctive, but then `INSEE` is just a number (for almost all values). This makes any consumer testing a field with numbers in that domain to match against this field easily.

image

In order to remove this source for false positives, a solution could be to completely avoid writing those values in our manifest by adding an optional property in the field mappings to opt them out.

```hjson
fieldMapping: [
...
{
    type: id
    name: insee
    desc: INSEE department identifier
    skipValues: true  <= new optional value
    alias: [
    insee
    ]
}
...
]
```

Then in the build process those marked fields would not contribute values to the corresponding manifest. 

I'm not sure if we should create new versioned source files with `versions: >= 8.2`, or we can introduce this change on the existing manifests.

jsanz:
Fixes #242

- Add a new `skipCopy` property for the `fieldMapping` that avoids copying actual field values into the manifest
- Add the new property to fields that are just numbers

To test this PR you can build the manifests and run `jq` on the world countries to check for fields of `type: id` and the existence of a `values` key.

For 8.1 it presents `iso_numeric`:

```
$ cat dist/vector/v8.1/manifest |\
  jq -c '.layers[] | 
    select (.layer_id == "world_countries") | 
    .fields[] |
    select(.type =="id" and .values != null) |
    { id : .id }'
{"id":"iso2"}
{"id":"iso3"}
{"id":"iso_numeric"}
```

And for `8.2` it does not present `iso_numeric`:

```
$ cat dist/vector/v8.2/manifest |\
  jq -c '.layers[] | 
    select (.layer_id == "world_countries") | 
    .fields[] |
    select(.type =="id" and .values != null) |
    { id : .id }'
{"id":"iso2"}
{"id":"iso3"}
```

The change is introduced in new source files from `8.2` leaving previous versions without change. The list of datasets modified are:

* Chile regions
* China provinces
* France departments
* Ireland counties
* Italy provinces
* Italy regions
* Japan prefectures
* US counties
* US zipcodes
* World Countries

:warning: **Note**: on the new source files I've removed the `regex` property where `skipCopy: true` as I understand the field is not necessary.

### DIFF
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -191,14 +191,19 @@ function getDefaultFormat(emsFormats) {
 }
 
 function getIdsFromFile(dataDir, file, fields) {
+  const fieldMap = {};
+  const fieldsWithIds = fields.filter(field => !field.skipCopy);
+
+  if (fieldsWithIds.length == 0) return fieldMap;
+
+  // Only read the dataset if there are identfiers to return
   const json = JSON.parse(readFileSync(`${dataDir}/${file}`, 'utf8'));
   const features = json.features || json.objects.data.geometries;
-  const fieldMap = {};
-  for (const { name } of fields) {
+  for (const { name } of fieldsWithIds) {
     fieldMap[name] = new Set(); // Probably unnecessary but ensures unique ids
   }
   for (const feature of features) {
-    for (const { name } of fields.filter( field => ! field.skipCopy)) {
+    for (const { name } of fieldsWithIds) {
       fieldMap[name].add(feature.properties[name]);
     }
   }
@@ -273,8 +278,8 @@ function getFieldMapping(sourceFieldsMap, manifestVersion, idInfos, fieldInfo) {
   return sourceFieldsMap
     .filter(sourceFieldMap => ['id', 'property'].includes(sourceFieldMap.type))
     .map(sourceFieldMap => {
-      const { type, name, desc, regex, alias } = sourceFieldMap;
-      const values = type === 'id' ? [...idInfos[name]] : undefined;
+      const { type, name, desc, regex, alias, skipCopy } = sourceFieldMap;
+      const values = type === 'id' && !skipCopy ? [...idInfos[name]] : undefined;
       return {
         type,
         id: name,
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -198,7 +198,7 @@ function getIdsFromFile(dataDir, file, fields) {
     fieldMap[name] = new Set(); // Probably unnecessary but ensures unique ids
   }
   for (const feature of features) {
-    for (const { name } of fields) {
+    for (const { name } of fields.filter( field => ! field.skipCopy)) {
       fieldMap[name].add(feature.properties[name]);
     }
   }
--- a/test/fixtures/valid-sources/sources.json
+++ b/test/fixtures/valid-sources/sources.json
@@ -244,5 +244,56 @@
       "default": false
     }],
     "createdAt": "1532-12-25T18:45:32.389979"
+  },
+  {
+    "versions": ">=7.14",
+    "production": false,
+    "attribution": [
+      {
+        "label": {
+          "en": "The Silmarillion"
+        }
+      }
+    ],
+    "type": "http",
+    "data": "http://example.com",
+    "name": "shireWithoutIds",
+    "legacyIds": [ "Shire", "Shire regions", "Shire Regions" ],
+    "humanReadableName": {
+      "en": "Shire regions",
+      "de": "Auenland",
+      "zh": "夏爾"
+    },
+    "fieldMapping": [
+      {
+        "type": "id",
+        "name": "wikidata",
+        "desc": "Wikidata identifier",
+        "skipCopy": true
+      },
+      {
+        "type": "property",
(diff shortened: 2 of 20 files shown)


test/manifest-v7.14.js
```diff
@@ -210,6 +210,66 @@ function getExpectedVector(version) {
...
+            type: 'topojson',
+            url: '/files/shire_v2.topo.json',
+            legacy_default: false,
+          },
+        ],
+        fields: [
+          {
+            type: 'id',
+            id: 'wikidata',
+            values: [],
```
jsanz, line 239: This is the empty `values` array that the `skipCopy` property should generate.

test/fixtures/valid-sources/sources.json
```diff
@@ -244,5 +244,56 @@
...
+      "en": "Shire regions",
+      "de": "Auenland",
+      "zh": "夏爾"
+    },
+    "fieldMapping": [
+      {
+        "type": "id",
+        "name": "wikidata",
+        "desc": "Wikidata identifier",
+        "skipCopy": true
```
jsanz, line 272: This is the new property that specifies that this identifier is made of common values that could produce false positives.

scripts/generate-manifest.js
```diff
@@ -198,7 +198,7 @@ function getIdsFromFile(dataDir, file, fields) {
     fieldMap[name] = new Set(); // Probably unnecessary but ensures unique ids
   }
   for (const feature of features) {
-    for (const { name } of fields) {
+    for (const { name } of fields.filter( field => ! field.skipCopy)) {
```
jsanz, line 201: Main logic introduced in this PR. This leaves an empty `new Set()` in the output identifier definition.

test/manifest-v7.14.js
thomasneirynck: iso empty, I'd consider removing this altogether. The field is optional anyway https://github.com/elastic/kibana/blob/e45209b49f6cfa275d346847e9176e1d8261c22a/x-pack/plugins/maps/public/ems_autosuggest/ems_autosuggest.ts#L28

thomasneirynck (approved):
thx! This should clear a lot of issues already 🤞

test/manifest-v7.14.js
jsanz: If it's definitely OK to return the object without the values I'll change the code to do so, it's cleaner.

test/manifest-v7.14.js
```diff
@@ -236,7 +236,6 @@ function getExpectedVector(version) {
           {
             type: 'id',
             id: 'wikidata',
-            values: [],
```
jsanz: The `values` property is not expected anymore for field definitions with `skipCopy` property set to `true`.

jsanz:
How to test this with Kibana.

I've uploaded a [distribution](https://storage.googleapis.com/jsanz-bucket/tmp/ems-file-service-pr243/) of the result of building EMS File Service with these changes so you can run the following test or similar:

* Run Kibana from `main`, be sure you checked changes after https://github.com/elastic/kibana/pull/128525
* Add the following data that takes the same values of the first page of [China provinces](https://maps.elastic.co/#file/china_provinces) `division_code` field from maps.elastic.co

```

PUT common_ids
{
  "mappings": {
    "properties": {
      "id": { "type": "keyword" }
    }
  },
  "settings": {
    "number_of_replicas": 0
  }
}

POST _bulk
{ "index" : { "_index" : "common_ids" } }
{ "id" : "50" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "12" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "82" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "44" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "45" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "32" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "54" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "23" }
{ "index" : { "_index" : "common_ids" } }
{ "id" : "51" }
```

* Create a Data View
* Open with the ML Data Viewer and expand the `id` field. You should see the false positive with a map of China provinces

common_ids_pre

* Now edit `kibana.dev.yml` and add the following line pointing to the custom distribution

```
map.emsFileApiUrl: https://storage.googleapis.com/jsanz-bucket/tmp/ems-file-service-pr243/vector
```

* Open again the ML Data Visualizer and the map should be gone

common_ids_post


I'm pretty sure you can test this with other fields from the list of updated datasets.

jsanz:
@nickpeihl @thomasneirynck this PR is against the `feature-layers` branch so this is not yet in production and can be checked as usual with

```
map.emsFileApiUrl: https://storage.googleapis.com/elastic-bekitzur-emsfiles-vector-dev/
```

Let me know if you think we should do further testing before pushing to production this upgrade.

PR #245, opened 2022-03-29
Fix DK-85

jsanz:
Fixes elastic/kibana#128721

- New NE regions dataset with DK-85 region name fixed
- Update generated admin regions datasets


nickpeihl (approved):
Lgtm!

PR #260, opened 2022-11-16
Fix Morocco regions human readable labels

### LINKED ISSUE #259: Fix Morocco regions label
The source file from #258 needs to capitalize the human labels at

https://github.com/elastic/ems-file-service/blob/master/sources/ma/regions_v1.hjson#L98-L138 

image

jsanz:
Fixes #259

Small fixes to human readable labels in different langagues to uppercase the initial and change the English lable to follow the same pattern as all the other layers with the name of the region first.

cc. @nelmalki


nickpeihl (approved):
lgtm

Thanks

PR #261, opened 2022-11-21
Update dependencies

jsanz:
Broad update of dependencies to more modern versions.

JSTS has been pinned at the last release before it was upgraded to a module. This will need further work, that in any case, may be convenient to modernize the entire codebase.


package.json
```diff
@@ -25,23 +25,19 @@
...
     "babel-eslint": "^10.1.0",
     "csv-parse": "^4.12.0",
-    "eslint": "^6.8.0",
+    "eslint": "^7.32.0",
     "eslint-plugin-babel": "^5.3.0",
     "eslint-plugin-import": "^2.20.2",
-    "eslint-plugin-jest": "^23.8.2",
-    "eslint-plugin-mocha": "^6.3.0",
     "eslint-plugin-prefer-object-spread": "^1.2.1",
-    "eslint-plugin-react": "^7.19.0",
```
nickpeihl, line 40: 💯  I think I originally just copied all the `eslint-*` dependencies from kibana even though we weren't using them. 😅

nickpeihl (approved):
lgtm 

code review only

package.json
jsanz: haha, no prob! :+1:

PR #263, opened 2023-01-24
Update versions for manifest generation and tests

jsanz:
Removing versions of the stack that won't be released and adding a few new future versions, similar to #229 and #252 previously.


nickpeihl (approved):
lgtm

code review

PR #266, opened 2023-02-16
Upgrade ajv, node-fetch, tap, eslint-plugin, and fix build

jsanz:
* Just a few upgrades run through `yarn upgrade-interactive`.
* Removes the `spin` config setting from `build.sh`
* Adds a `--no-timeout` option to `tap` to prevent tests to fail


nickpeihl (approved):
lgtm! 

code review

PR #271, opened 2023-04-04
Deploying from buildkite

jsanz:
**WIP**

Adding new steps on the Buildkite pipeline definition to deploy EMS File Service assets in GCP bucket.

:warning: At this moment this is getting credentials and deploying in the EMS development GCP project. This will be updated with production credentiars and bucket.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,8 +2,26 @@ agents:
   image: "docker.elastic.co/ci-agent-images/ems/buildkite-agent:latest"
 
 steps:
-  - label: ":hammer_and_wrench: Test and build"
+  - key: test
+    label: ":hammer_and_wrench: Test and build"
     commands: 
       - "yarn install"
       - "yarn test"
-      - "yarn build"
+    artifact_paths: "dist/*"
+
+  - wait:
+      allow_dependency_failure: false
+
+  - key: deploy-dev
+    label: ":rocket: Deploy to dev"
+    # branches: "feature-layers"
+    depends_on: test
+    command: ".buildkite/deploy.sh"
+    env:
+      GPROJECT: "elastic-ems-dev"
+      GCS_VAULT_SECRET_PATH: "secret/ci/elastic-ems-file-service/testing/gcs_acount"
+      TILE_HOST: "tiles.maps.elastic.co"
+      CATALOGUE_BUCKET: "elastic-ems-dev-emsfiles-catalogue-dev"
+      VECTOR_BUCKET: "elastic-ems-dev-emsfiles-vector-dev"
+      VECTOR_HOST: "storage.googleapis.com/elastic-ems-dev-emsfiles-vector-dev"
+
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,7 +8,8 @@ steps:
       - "yarn install"
       - "yarn test"
       - "yarn build"
-    artifact_paths: "dist/*"
+      - "tar cf dist.tar dist"
+    artifact_paths: "dist.tar.gz"
 
   - wait:
       allow_dependency_failure: false
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,6 +7,7 @@ steps:
     commands: 
       - "yarn install"
       - "yarn test"
+      - "yarn build"
     artifact_paths: "dist/*"
 
   - wait:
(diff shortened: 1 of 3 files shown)


jsanz:
@nickpeihl I thought maybe some documentation of how things work now could be helpful for your review, no problem on rewriting whatever we change. While writing I was thinking maybe of a couple of changes that we may want regarding notifications.

Should we include a Slack mention to our team when running builds that trigger deployments? so we get a bit more of attention to those? Maybe for this case it is not necessary since the builds run pretty fast.

Additionally, should we send also emails to the mailing list we have for operations when builds fail? Again, maybe it is too much noise for something that we see quickly how it went, and it makes more sense for builds that take more time or that run on a schedule.

jsanz:
Just another comment @nickpeihl, maybe we should add on this PR an `exit 0` to the beggining of the scripts running from Jenkins to convert them into dumb executions so we don't deploy twice. And once we are certain Buildkite is running correctly then remove all the Jenkins infrastructure in a later PR (and request to our infra team). Makes sense?

nickpeihl:
> Just another comment @nickpeihl, maybe we should add on this PR an `exit 0` to the beggining of the scripts running from Jenkins to convert them into dumb executions so we don't deploy twice. And once we are certain Buildkite is running correctly then remove all the Jenkins infrastructure in a later PR (and request to our infra team). Makes sense?

Makes sense to me. We definitely don't want Jenkins and Buildkite simultaneously writing to the same bucket.

.buildkite/pipeline.yml
```diff
@@ -2,8 +2,37 @@ agents:
...
+  - key: deploy-staging
+    label: ":rocket: Deploy to the staging environment"
+    branches: "master"
+    depends_on: test
+    command: ".buildkite/deploy.sh"
+    env:
+      EMS_ENVIRONMENT: "staging"
+
+  - key: should-deploy
+    block: ":one-does-not-simply: Deploy to production?"
```
nickpeihl, line 28: `:one-does-not-simply:` 😆

README.md
```diff
@@ -26,10 +26,10 @@ yarn build
...
 - Pull requests for new feature layers should be made from the `feature-layers` against the `master` branch. Pull requests for any other changes should be made on a new branch in your fork, e.g. `git checkout -b my-bugfix`.
-- Once merged, Jenkins will run `deployStaging.sh` script, which will place the contents of the `dist` directory into the staging bucket.
-- Deploying to production requires manually triggering [this Jenkins job](https://kibana-ci.elastic.co/job/elastic+ems-file-service+deploy/) to run the `deployProduction.sh` script. This will rsync files from the staging bucket to the production bucket. To trigger, log in and click the "Build with Parameters" link. Leave the `branch_specifier` field as default (`refs/heads/master`).
+- Once merged, Buildkite will run `.buildkite/deploy.sh` script, which will place the contents of the `dist` directory into the staging bucket.
+- Deploying to production requires pushing a tag  (`git push --tags`) and accepting the `Deploy to production?` block step. This will execute the `deploy.sh` script against the production GCP buckets, creating also an archive in a separate bucket for future reference.
```
nickpeihl, line 32: Should we standardize on a tag format? If so, maybe we should list that format here.

Should the format should be semantic versioning like `v<major>.<minor>.<patch>`? This is how [we tag Kibana](https://github.com/elastic/kibana/tags).

nickpeihl (approved):
lgtm! I just added one concern. code and documentation review only. 🤞

nickpeihl:
> Should we include a Slack mention to our team when running builds that trigger deployments? so we get a bit more of attention to those? Maybe for this case it is not necessary since the builds run pretty fast.

Yes, I think that's a good idea. We can always remove it later if we get too annoyed. 🙈 

> Additionally, should we send also emails to the mailing list we have for operations when builds fail? Again, maybe it is too much noise for something that we see quickly how it went, and it makes more sense for builds that take more time or that run on a schedule.

I honestly don't use the mailing list for checking builds. I only look at #maps-ops (when I think about it 😅). I think it would be nice to get notifications on build failures. Again, we can remove them later if they generate too much noise.

README.md
jsanz: I was thinking on a date approach tag scheme, to avoid confusion with the versions supported by our generation script and driven by the `VERSIONS` array at `scripts/contants.js`.

What do you think?

jsanz:
OK so I will configure a mention but we can leave the mailing list aside for the moment. I will ping you to another review for these minor changes.

jsanz:
@nickpeihl I've added a few small changes after feedback, please let me know if there's anything to improve

https://github.com/elastic/ems-file-service/compare/00c1bc3...d1ff7ee

PR #273, opened 2023-04-18
Fix archive bucket url

jsanz:
Fixes the url for the archive bucket location on the production build environment.


nickpeihl (approved):
🙈 Oops! I didn't see that in my last review! Lgtm!

PR #277, opened 2023-04-19
Remove Jenkins CI jobs and scripts

jsanz:
After testing Builkite executions on our three environments we can remove all the code from Jenkins.

Webhooks have been already removed and a follow up PR will be created to remove the jobs in the Elastic infrastructure.


nickpeihl (commented):
I might be misunderstanding how this works, but I think we need two separate PRs to accomplish this.

PR 1 would only delete the `elastic+ems-file-service+.yml` files. Once those are deleted, the CI will remove them from the job registry.
PR 2 would delete the rest of the `.ci` directory including `check_paths_for_matches`. 

Maybe @jonahbull can correct me if I'm wrong?

jonahbull:
@nickpeihl Yep, that's correct, we had some of that discussion in https://github.com/elastic/infra/pull/38865

jsanz:
Let's start again :sweat_smile:

PR #278, opened 2023-04-20
Remove Jenkins jobs

jsanz:
Replaces #277

First PR to remove Jenkins integration by deleting just the job definitions.

PR #279, opened 2023-04-20
Remove deploy scripts and .ci folder

jsanz:
Follow up from #278, after getting the jobs removed from our infra CI, we can remove the rest of the Jenkins related artifacts.


nickpeihl (approved):
lgtm! I think all the jobs have now been deleted from ci, so this should just finalize the clean-up

PR #287, opened 2023-07-14
Support date versions

jsanz:
Resolves #286

This PR introduces support to define release versions tagged as ISO8601 (yyyy-mm-yy). A new array called `DATE_VERSIONS` in the `constants.js` file contains objects as in the block below. This definition provides the minimum changes to how manifests are generated, matching a date release with a Semantic release. Finally, a "latest" folder is also generated with the last entry on the `DATE_VERSIONS` array.

```
  ...
  DATE_VERSIONS: [
    {
      date: "2023-10-31",
      semver: "v8.10",
    },
  ],
  LATEST_TAG: "latest",
  ...
```

PR #307, opened 2024-03-06
Update README with versioning details

jsanz:
Just adding some details to the `README` about how we manage date-based versions, to give more details to #287.


nickpeihl (approved):
lgtm!

PR #311, opened 2024-05-17
ESM upgrade

jsanz:
I wanted to upgrade dependencies in this repo and started shaving a yak :sweat_smile:. I ended up having to upgrade from CommonJS to ESM to upgrade JSTS.

* General upgrade of all dependencies
* Upgraded code to ESM modules, replacing the `require` statements by `import`.
* For JSTS, this upgrade meant that the `isSimple`, `isValid`, and `buffer` methods needed to be imported separately and used as class instances that received the geometry and applied that operation.
* Upgraded `eslint`: removing old style configuration files in favor of a flat `eslint.config.js` file.
* Upgrading `lodash` to `lodash-es` and getting the required functions as individual imports instead of the global underscore.
* `mapshaper` does not provide a `rf7946` output option anymore, using the `precision` parameter, yielding the exact same outputs.

> [!note]
> I've left a few comments in the files to try to give more details on the changes.


scripts/layerdiff.js
```diff
@@ -76,10 +80,9 @@ const getId = function (f) {
 };
 
 const toFeaturePoint = function (geom) {
-  const { x, y } = geom.getCoordinates()[0];
```
jsanz, line 79: this method was not available on the updated API, using `getX` and `getY` instead.

scripts/layerdiff.js
```diff
@@ -125,22 +128,25 @@ const compareFeatures = function ({ id, left, right }) {
...
-    toFeaturePoint(lCentroid),
-    toFeaturePoint(rGeom.getCentroid()), 'radians'));
+  const centroidDist = Math.floor(
+    getDistance(toFeaturePoint(lCentroid), toFeaturePoint(rCentroid))
+  );
 
   diffs.centroid = centroidDist > CENTROID_DIST;
   details.centroid = centroidDist;
 
-  details.centroids = { left: lCentroid.toText(), right: rCentroid.toText() };
```
jsanz, line 143: `toText` was not available on the updated API, using `getX` and `getY` instead.

scripts/wikidata-labels.js
```diff
@@ -8,7 +8,6 @@ if (!wdItem) {
 }
 
 async function wikidataTranslations(id) {
-  try {
```
jsanz, line 11: `eslint` did not like this try/catch for a good reason :sweat_smile:

sources/world/Makefile
```diff
@@ -47,21 +47,21 @@ clean: ## Remove output files
 ../../data/admin_regions_lvl2_v2.geo.json: ../../data/admin_regions_lvl2_v2.topo.json ## Regions GeoJSON full size
 	mapshaper -i $(DATA_DIR)/admin_regions_lvl2_v2.topo.json \
   -sort this.properties.region_iso_code \
-  -o rfc7946 id-field=region_iso_code format=geojson force /tmp/admin_regions_lvl2_v2.geo.json
```
jsanz, line 50: In [mapshaper 0.5.6](https://github.com/mbloch/mapshaper/releases/tag/v0.5.6) there was this change:

> The old rfc7946 flag is no longer useful, but is retained for backwards compatibility. This flag (as before) truncates coordinates to 6 decimal places. Now, using this flag is equivalent to adding precision=0.000001 to the output command.

test/manifest-date.js
```diff
@@ -4,23 +4,28 @@
...
   coerceToDateSemver,
-} = require("../scripts/date-versions");
-
-const sources = require("./fixtures/valid-sources/sources.json");
-const fieldInfo = require("./fixtures/fieldInfo.json");
-const constants = require("../scripts/constants");
+} from "../scripts/date-versions.js";
+import constants from "../scripts/constants.js";
+
+const sources = JSON.parse(
```
jsanz, line 23: I did not find a better way to import these JSON files. Is not super clean but :shrug:

sources/world/Makefile
nickpeihl: Thanks for fixing this. I don't agree with Mapshaper that `precision=0.000001` is better than `rfc7946` because it is more verbose and more error prone (easy to introduce typos). But that's just my opinion and we should use their preferred flag.

test/manifest-date.js
nickpeihl: AFAIK, that's the best way to import JSON.

test/source.js
```diff
@@ -67,11 +77,13 @@ function testSourceFiles(source) {
 
 function validateObjectsMember(topojson, format, fieldMap, t) {
   const fc = JSON.parse(topojson);
-  const fcPath = _.get(format, 'meta.feature_collection_path', 'data');
+  const fcPath = get(format, "meta.feature_collection_path", "data");
   const fieldsNames = fieldMap.map(f => f.name).sort();
+  //eslint-disable-next-line no-prototype-builtins
   t.ok(fc.objects.hasOwnProperty(fcPath));
```
nickpeihl, line 83: I think we can just [make lint happy](https://eslint.org/docs/latest/rules/no-prototype-builtins) by changing this.

```suggestion
  t.ok(Object.prototype.hasOwnProperty.call(fc.objects, fcPath));
```

test/source.js
```diff
@@ -67,11 +77,13 @@ function testSourceFiles(source) {
...
   t.type(fc.objects[fcPath], 'object');
 
+  //eslint-disable-next-line no-prototype-builtins
   t.ok(fc.objects[fcPath].hasOwnProperty('geometries'));
```
nickpeihl, line 87: nit

```suggestion
  t.ok(Object.prototype.hasOwnProperty.call(fc.objects[fcPath], 'geometries'));
```

index.js
```diff
@@ -78,6 +80,7 @@ for (const version of versions) {
 for (const file of vectorFiles) {
   // file is an array of [dest, src]
   const vector = JSON.parse(fs.readFileSync(file[1]));
+  //eslint-disable-next-line no-prototype-builtins
   const vectorToWrite = vector.hasOwnProperty('type')
```
nickpeihl, line 84: nit

```suggestion
  const vectorToWrite = Object.prototype.hasOwnProperty.call(vector, 'type')
```

nickpeihl (approved):
lgtm! just added a couple of nits.

sources/world/Makefile
jsanz: I agree, it makes more sense to me to be explicit about the standard. The way they deprecated it was that the flag did not raise an error, but also had no effect at all! :shrug:

scripts/layerdiff.js
```diff
@@ -251,6 +257,7 @@ const splitFeatures = function (leftJson, rightJson) {
 
   // Find if there are any features from the left
   // without the right side
+  //eslint-disable-next-line no-unsafe-negation
   const notRight = features.filter(f => !'right' in f);
```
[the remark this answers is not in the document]
jsanz: Yeah, I wanted to avoid touching even more code, leaving those for later, but it makes sense to fix them here as well. Thanks!!

jsanz:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|feature-layers|[](https://github.com/elastic/ems-file-service/pull/312)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #750, opened 2026-08-28
chore(ci): parallelize tap test files to speed up CI

jsanz:
## Summary

- Adds `--jobs 4` to the tap test command as a baseline measurement
- Follow-up commit will split `test/source.js` into `source-world.js` and `source-us.js` to allow the slow geometry validation tests to run in parallel

## Test plan

- [ ] Measure CI wall-clock time on this commit (baseline: `--jobs 4` with single `source.js`)
- [ ] Measure CI wall-clock time after file split commit
- [ ] Compare: expected ~100s → ~40s improvement for test suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint --fix *.js scripts/*.js",
-    "test": "tap --timeout=0 test/*.js",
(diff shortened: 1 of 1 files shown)


jsanz:
buildkite test this

